Issue #265, opened 2022-07-25
Feat: Change justification tokens to be struct rather than string. 


raserva:
Setting as string is not query friendly. I think we should convert it a `*structpb.Value`, see example: https://github.com/abcxyz/lumberjack/blob/main/clients/go/pkg/audit/runtime.go#L251

_Originally posted by @yolocs in https://github.com/abcxyz/lumberjack/pull/261#discussion_r929354662_

yolocs:
Still need to update java client

Issue #266, opened 2022-07-27
Populate standard audit log field for justification


yolocs:
The field being https://github.com/googleapis/googleapis/blob/master/google/rpc/context/attribute_context.proto#L199-L201

`log.RequestMetadata.RequestAttributes.Reason`

We currently add the justification token to the log metadata. But we should consider also populating the standard `reason` field with the text justification.

cc: @raserva

raserva:
I agree that it could be beneficial to copy the "explanation" field from the justificaiton token to the reason field.

yolocs:
Additional idea: Justification can now be auto populated with gRPC interceptor. But when a user doesn't use the interceptor (e.g. not a gRPC service), the parsing/populating of the justification becomes harder. We can add a lib that 1) verify/parse a given justification token, 2) populate a audit log request with the justification

yolocs:
This becomes more important as some downstream component expecting the justification in `log.RequestMetadata.RequestAttributes.Reason`.
We can json encode the `justs` claim in the JVS token and put that in this field.

capri-xiyue:
golang part https://github.com/abcxyz/lumberjack/pull/293 completed

PR #272, opened 2022-08-01
feat: Go client - add util lib to populate justification info in audit log

yolocs:
And use the util lib in the interceptor. The motivation is to enable non-grpc users to use the util lib to populate audit log with justification info.

**Notice:** I had to change the key name in the log metadata because previously we used `string` value and it became a part of BigQuery table schema. I can't easily change column type for nested columns (aka. `RECORD`) so I have to use a new metadata key. Before we make the same change in java client, I need to temporarily modify the integration test to check justification in both old/new log metadata.

Context in: https://github.com/abcxyz/lumberjack/issues/266
Also fixes: https://github.com/abcxyz/lumberjack/issues/265


clients/go/pkg/justification/processor.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	b, err := json.Marshal(*tok)
+	if err != nil {
+		return fmt.Errorf("failed to encode justification token: %w", err)
+	}
+
+	tokStruct := &structpb.Struct{}
+	if err := protojson.Unmarshal(b, tokStruct); err != nil {
+		return fmt.Errorf("failed to encode justification token: %w", err)
```
capri-xiyue, line 60: "failed to decode justification token"?

clients/go/pkg/audit/interceptor.go
```diff
@@ -329,6 +262,30 @@ func (i *Interceptor) StreamInterceptor(srv interface{}, ss grpc.ServerStream, i
...
+	if i.justProcessor == nil {
+		return nil
+	}
+
+	// Justification processor is present, look for justification info
+	// TODO(#257): If we add justification mode, we need to honor that.
+	jvsToken := ""
+	md, ok := grpcmetadata.FromIncomingContext(ctx)
+	if ok {
+		vals := md.Get("justification-token")
```
sethvargo, line 277: Nit: const

clients/go/pkg/justification/processor.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// Process populates the given audit log request with the justification info from the given token.
+// If the token is empty, this function does nothing.
+func (p *Processor) Process(jvsToken string, logReq *api.AuditLogRequest) error {
+	if jvsToken == "" {
+		return nil
+	}
+
+	tok, err := p.Validator.ValidateJWT(jvsToken)
```
sethvargo, line 48: Since `Validator` is a public field, it's possible for it to be accidentally set to `nil` and then this will NPE.

clients/go/pkg/justification/processor.go
```diff
@@ -0,0 +1,76 @@
...
+		return fmt.Errorf("failed to validate justification token: %w", err)
+	}
+
+	b, err := json.Marshal(*tok)
+	if err != nil {
+		return fmt.Errorf("failed to encode justification token: %w", err)
+	}
+
+	tokStruct := &structpb.Struct{}
+	if err := protojson.Unmarshal(b, tokStruct); err != nil {
```
sethvargo, line 59:
```suggestion
	var tokStruct structpb.Struct
	if err := protojson.Unmarshal(b, &tokStruct); err != nil {
```

clients/go/pkg/justification/processor.go
```diff
@@ -0,0 +1,76 @@
...
+		return fmt.Errorf("log request missing payload")
+	}
+
+	if logReq.Payload.Metadata == nil {
+		logReq.Payload.Metadata = &structpb.Struct{
+			Fields: map[string]*structpb.Value{},
+		}
+	}
+
+	logReq.Payload.Metadata.Fields["justification_token"] = structpb.NewStructValue(tokStruct)
```
sethvargo, line 73: Nit: const

clients/go/pkg/audit/interceptor.go
```diff
@@ -82,28 +82,21 @@ func WithInterceptorLogMode(m api.AuditLogRequest_LogMode) InterceptorOption {
...
 type Interceptor struct {
 	*Client
-	sc           security.GRPCContext
-	rules        []*api.AuditRule
-	logMode      api.AuditLogRequest_LogMode
-	jwtValidator JWTValidator
+	sc            security.GRPCContext
+	rules         []*api.AuditRule
+	logMode       api.AuditLogRequest_LogMode
+	justProcessor *justification.Processor
```
sethvargo, line 99: Nit: shortened names this are a bit weird. Can we just say "justificationProcessor"?

raserva (commented):
I like the change! other than the issues called out by seth/xiyue, LGTM

clients/go/pkg/audit/interceptor.go
capri-xiyue: +1

sethvargo (approved):

capri-xiyue (approved):

PR #273, opened 2022-08-03
feat: Java client - add util lib to validate and fill justification in logs

yolocs:
Fixes: https://github.com/abcxyz/lumberjack/issues/265
Similar to Go client change in: https://github.com/abcxyz/lumberjack/pull/272

My formatter did a bunch of changes and I don't know if it's the proper. I think my VS code is using https://raw.githubusercontent.com/google/styleguide/gh-pages/eclipse-java-google-style.xml

p.s. This is my first java code change in like 7 years.

### DIFF
--- a/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
+++ b/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
@@ -1,30 +1,32 @@
 /*
  * Copyright 2022 Lumberjack authors (see AUTHORS file)
  *
- * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
- * in compliance with the License. You may obtain a copy of the License at
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software distributed under the License
- * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
- * or implied. See the License for the specific language governing permissions and limitations under
- * the License.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  */
 
 package com.abcxyz.lumberjack.auditlogclient.processor;
 
 import com.abcxyz.jvs.JvsClient;
 import com.abcxyz.lumberjack.v1alpha1.AuditLogRequest;
+import com.auth0.jwt.interfaces.DecodedJWT;
 import com.google.api.client.util.StringUtils;
 import com.google.cloud.audit.AuditLog;
-import com.auth0.jwt.interfaces.DecodedJWT;
-import org.apache.commons.codec.binary.Base64;
 import com.google.inject.Inject;
 import com.google.protobuf.Struct;
 import com.google.protobuf.Value;
 import com.google.protobuf.util.JsonFormat;
 import lombok.AllArgsConstructor;
+import org.apache.commons.codec.binary.Base64;
 
 @AllArgsConstructor(onConstructor = @__({@Inject}))
 public class JustificationProcessor {
@@ -54,8 +56,8 @@ public class JustificationProcessor {
   }
 
   /**
-   * Validates the given {@code jvsToken} and sets the justification info in the given
-   * {@code auditLogBuilder}.
+   * Validates the given {@code jvsToken} and sets the justification info in the given {@code
+   * auditLogBuilder}.
    *
    * @param jvsToken A JWT produced by JVS.
    * @param auditLogBuilder Audit log builder.
@@ -77,7 +79,8 @@ public class JustificationProcessor {
 
       // add new field with monitoredResource to existing metadata
       Struct.Builder metadataBuilder = auditLogBuilder.getMetadata().toBuilder();
-      metadataBuilder.putFields(JUSTIFICATION_LOG_METADATA_KEY,
+      metadataBuilder.putFields(
+          JUSTIFICATION_LOG_METADATA_KEY,
           Value.newBuilder().setStructValue(justificationStruct).build());
       auditLogBuilder.setMetadata(metadataBuilder.build());
 
--- a/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
+++ b/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
@@ -18,10 +18,14 @@ package com.abcxyz.lumberjack.auditlogclient.processor;
 
 import com.abcxyz.jvs.JvsClient;
 import com.abcxyz.lumberjack.v1alpha1.AuditLogRequest;
+import com.auth0.jwk.JwkException;
 import com.auth0.jwt.interfaces.DecodedJWT;
+import com.google.api.client.util.Preconditions;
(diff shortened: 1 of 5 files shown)


raserva:
yeah, is it possible to undo the formatting changes? It looks like the java linter here isn't happy, and it makes it more difficult to review the specific changes you implemented.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+   * @param jvsToken A JWT produced by JVS.
+   * @param auditLogRequest Audit log request to be processed.
+   * @return The {@link AuditLogRequest} with justification filled.
+   * @throws LogProcessingException when it fails to populate the justification info.
+   */
+  public AuditLogRequest process(String jvsToken, AuditLogRequest auditLogRequest)
+      throws LogProcessingException {
+    if (!auditLogRequest.hasPayload()) {
+      throw new IllegalArgumentException("audit log request doesn't have payload");
+    }
```
raserva, line 47:
```suggestion
    Preconditions.checkArgument(auditLogRequest.hasPayload(), "audit log request doesn't have payload");
```

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Validates the given {@code jvsToken} and sets the justification info in the given
+   * {@code auditLogBuilder}.
+   *
+   * @param jvsToken A JWT produced by JVS.
+   * @param auditLogBuilder Audit log builder.
+   * @throws LogProcessingException when it fails to set the justification info.
+   */
+  public void setLogJustification(String jvsToken, AuditLog.Builder auditLogBuilder)
```
raserva, line 64: It is atypical within java to have a method that modifies the inputs. Usually any results should be returned. Therefore, the more idiomatic way would be to have this return the metadata, and then allow the calling function to set it in the AuditLog.Builder

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+      throws LogProcessingException {
+    try {
+      DecodedJWT jwt = jvs.validateJWT(jvsToken);
+      String jsonString = StringUtils.newStringUtf8(Base64.decodeBase64(jwt.getPayload()));
+      Struct.Builder justificationStructBuilder = Struct.newBuilder();
+      JsonFormat.parser().merge(jsonString, justificationStructBuilder);
+      Struct justificationStruct = justificationStructBuilder.build();
+
+      // Add monitored resource to Payload.Metadata as JSON.
```
raserva, line 73: This comment doesn't seem to line up with what is happening (which is setting a blank metadata if there is none)

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+        auditLogBuilder.setMetadata(Struct.newBuilder().build());
+      }
+
+      // add new field with monitoredResource to existing metadata
+      Struct.Builder metadataBuilder = auditLogBuilder.getMetadata().toBuilder();
+      metadataBuilder.putFields(JUSTIFICATION_LOG_METADATA_KEY,
+          Value.newBuilder().setStructValue(justificationStruct).build());
+      auditLogBuilder.setMetadata(metadataBuilder.build());
+
+    } catch (Exception e) {
```
raserva, line 84: This catch clause is very broad. Can we make this more specific?

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.commons.codec.binary.Base64;
+import com.google.inject.Inject;
+import com.google.protobuf.Struct;
+import com.google.protobuf.Value;
+import com.google.protobuf.util.JsonFormat;
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor(onConstructor = @__({@Inject}))
+public class JustificationProcessor {
+  private static final String JUSTIFICATION_LOG_METADATA_KEY = "justification";
```
raserva, line 31: it is justification_token in golang, is it not? https://github.com/abcxyz/lumberjack/commit/a76a124e56323b48128f804199d9c0903dc9abf4#diff-60c8ed893b3d6f4959e9200c588be5b2df2ae81255d006f0b17e718e307232cdR46

raserva (changes requested):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
yolocs: That's _exactly_ what I'm trying to avoid. I don't want the caller to worry about where to set the justification.
Maybe I should return the same `auditLogBuilder`?

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
yolocs: Removed.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
yolocs: I changed it per https://github.com/abcxyz/lumberjack/pull/272. See reason in that PR's description.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
raserva: I believe that this should take a string and return a struct. One property of java is that it does not create copies of arguments passed in, so modifying an arugment modifies the underlying object and all future references to that object, even outside the method. Therefore, it has become standard practice to either explictly make or treat inputs as final if possible, and return modified values.

raserva: Discussed offline with @myurtoglu  who has Java readability, he seemed to be on the same page.

yolocs: Sure. But that's against my original purpose, which is to make a lib to properly set the justification in the _right_ place in the log. How should achieve that?

raserva: So the process method is fine. I think the process method should take the output of this method (which will return a struct) and set it in the audit log.

raserva: So, here https://github.com/abcxyz/lumberjack/pull/273/files#diff-ec13edea722c4098caa7f8b168cb91a5701fdeeb83f84a43cd3d1895b8d589adR56 take the struct and set it in the auditlog builder.

yolocs: I thought about that. But then in https://github.com/abcxyz/lumberjack/pull/273/files#diff-1386f01992f8e23e4621a9698030e5661cbc9368288e50a4018158d86d27820e, it only gives an `auditLogBuilder` to set justification. How should I handle that?

yolocs: As discussed offline, I changed the signature to return a copy of the builder. Thanks for the help!

raserva (approved):

PR #275, opened 2022-08-15
doc: break down and refine docs

yolocs:
- Added Go examples but need a little help on Java examples (will follow up).
- Pending less user facing docs (testing and setup)

### DIFF
--- a/docs/clients.md
+++ b/docs/clients.md
@@ -4,8 +4,8 @@
 
 Lumberjack client consists of the following processors and they run in order:
 
-1.  Validator(s): To validate audit log requests
-2.  Mutator(s): To mutate audit log request, e.g. applying common labels
+1.  Validator(s): To validate the audit log requests
+2.  Mutator(s): To mutate the audit log requests, e.g. applying common labels
 3.  Backend(s): Specifying where to send the logs.
 
 You can use a [config file](./config.md) to ease the client initialization or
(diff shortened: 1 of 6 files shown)


docs/clients.md
```diff
@@ -0,0 +1,100 @@
+# Clients
+
+**Lumberjack is not an official Google product.**
+
+Lumberjack client consists of the following processors and they run in order:
+
+1.  Validator(s): To validate audit log requests
+2.  Mutator(s): To mutate audit log request, e.g. applying common labels
```
myurtoglu, line 8: Maybe adding a "the" here to make it "To mutate the audit log request," is a good idea.

docs/clients.md
```diff
@@ -0,0 +1,100 @@
...
+  regex:
+    principal_include: "@example1\\.com$|@example2\\.com$"
+backend:
+  cloudlogging:
+    default_project: true
+labels:
+  common_label_1: foobar
+```
+
+### Exend
```
myurtoglu, line 69: Extend?

docs/config.md
```diff
@@ -0,0 +1,139 @@
...
+
+Please refer to the [gRPC guide](./grpc.md).
+
+## (WIP) Justification
+
+TODO
+
+## Supported env vars
+
+| ENV VAR name                                      | Description              |
```
myurtoglu, line 90: This table is looking a bit weird in github code review's preview, hopefully it will render ok?

myurtoglu (approved):

docs/config.md
yolocs: It seems github markdown doesn't recognize this table syntax. Fixed now.

PR #279, opened 2022-08-23
docs: added jvs docs

### LINKED ISSUE #278: add jvs integration docs
### TL;DR

added jvs integration docs

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

capri-xiyue:
fix #278


myurtoglu (changes requested):

docs/config.md
```diff
@@ -81,9 +81,23 @@ labels:
 
 Please refer to the [gRPC guide](./grpc.md).
 
-## (WIP) Justification
+## Justification
 
-TODO
+By default, the client will use `localhost:8080` as the endpoint to retrieve public keys from the JVS.
+These public keys will be used to validate JWT tokens that are passed in through the "justification_token" header.
```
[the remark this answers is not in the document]
capri-xiyue: done

myurtoglu (approved):

docs/config.md
```diff
@@ -81,9 +81,23 @@ labels:
 
 Please refer to the [gRPC guide](./grpc.md).
 
-## (WIP) Justification
+## Justification
```
yolocs, line 84: It comes to be that we should group justification configs. E.g.

```yaml
justification:
  enabled: true
  jvs_jwks_endpoint: https://example.com. # this should be a required field if justification is enabled
```

yolocs: We also need to point out this is only meaningful for gRPC services (not http)

docs/config.md
```diff
@@ -81,9 +81,23 @@ labels:
...
 
-TODO
+By default, the client will use `localhost:8080` as the endpoint to retrieve public keys from the JVS.
```
yolocs, line 86: I think we should also describe how to use https://github.com/abcxyz/lumberjack/blob/main/clients/go/pkg/justification/processor.go in https://github.com/abcxyz/lumberjack/blob/main/docs/clients.md

yolocs (changes requested):

docs/config.md
yolocs: Actually since it's gRPC only, we should move this to https://github.com/abcxyz/lumberjack/blob/main/docs/grpc.md

capri-xiyue: > It comes to me that we should group justification configs. E.g.
> 
> ```yaml
> justification:
>   enabled: true
>   jvs_jwks_endpoint: https://example.com. # this should be a required field if justification is enabled
> ```

https://github.com/abcxyz/lumberjack/blob/d4bc50031aec168f9fdcf23d4d7217fc8ed0c4fb/clients/go/apis/v1alpha1/config_test.go#L58
This is not what the code does right now, do we want to change it?

yolocs: I would prefer changing it now. We can punt this PR until we make that change. In fact, I think we should fix the justification processor as we discussed before writing the doc.

capri-xiyue:
Hold until we finish https://github.com/abcxyz/lumberjack/issues/280 and https://github.com/abcxyz/lumberjack/issues/283

docs/config.md
capri-xiyue: done. Still change it here instead of https://github.com/abcxyz/lumberjack/blob/main/docs/grpc.md since it's no longer grpc only after merge of #283

docs/config.md
capri-xiyue: done

capri-xiyue:
> Hold until we finish #280 and #283

@yolocs, this PR is ready for another round of review.

docs/clients.md
```diff
@@ -32,19 +32,27 @@ if err != nil {
...
 if err != nil {
   // Handle err
 }
-l := &audit.LabelProcessor{DefaultLabels: map[string]string{
+lp := &audit.LabelProcessor{DefaultLabels: map[string]string{
   "common_label_1": "foobar",
 }}
 
-client, err = audit.NewClient(audit.WithValidator(m), audit.WithBackend(p), audit.WithMutator(l))
+// Create JVS Client
```
yolocs, line 48:
```suggestion
// If justification is required, create a JVS client.
```

yolocs (approved):
/lgtm

Issue #280, opened 2022-08-26
combine jvs related config into one object( java & golang)


capri-xiyue:
### TL;DR

combine jvs related config into one object for **_both Java and Golang_**

### Detailed design
example PR(Golang): https://github.com/abcxyz/lumberjack/pull/284

_No response_

yolocs:
Can this be closed? @capri-xiyue

capri-xiyue:
> Can this be closed? @capri-xiyue

missing java part

yolocs:
cc: @sqin2019

Issue #283, opened 2022-08-31
Refactor justification processor


yolocs:
To process JVS token, we current have:

- Go: https://github.com/abcxyz/lumberjack/blob/main/clients/go/pkg/justification/processor.go
- Java: https://github.com/abcxyz/lumberjack/blob/main/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java

Because we need to pass _per-call_ JVS token, they can't implement the common processor interfaces:

- Go: https://github.com/abcxyz/lumberjack/blob/main/clients/go/pkg/audit/client.go#L42
- Java: https://github.com/abcxyz/lumberjack/blob/main/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/LogProcessor.java

As a result, we cannot make them a part of processing in the audit client. The justification processor needs to be separately initialized from config in non-gRPC cases, which impacts user experience a lot.

To improve, in https://github.com/abcxyz/lumberjack/pull/282 we have added a new context field to the `AuditLogRequest` proto. This allows us to pass the justification token with this new field. Thus, we can refactor the justification processor the implement the common processor interface (see above).

In addition, we need to refactor the audit client initialization logic to initialize justification processor as part of the audit client.

PR #284, opened 2022-08-31
refactor justificiation config(golang)

capri-xiyue:
fix golang part of  #280

### DIFF
--- a/clients/go/apis/v1alpha1/config.go
+++ b/clients/go/apis/v1alpha1/config.go
@@ -352,21 +352,23 @@ func (r *AuditRule) SetDefault() {
 
 // Justification specifies the config used to integrate with JVS.
 type Justification struct {
-	// JVSPublicKeysEndpoint is the endpoint where public keys may be retrieved from the JVS.
+	// PublicKeysEndpoint is the endpoint where public keys may be retrieved from the JVS.
 	// These will be used to validate JWT tokens that are passed in through the
 	// "justification-token" header.
-	JVSPublicKeysEndpoint string `yaml:"jvs_public_keys_endpoint,omitempty"`
+	PublicKeysEndpoint string `yaml:"public_keys_endpoint,omitempty" env:"JUSTIFICATION_PUBLIC_KEYS_ENDPOINT,overwrite"`
 
 	// Enabled indicates whether enables adding justification information to audit logs or not. If this is enabled,
 	// all manual calls are expected to pass in a justification in the "justification-token" header.
 	// If omitted, justifications will not be added to logs, even if provided.
-	Enabled bool `yaml:"enabled,omitempty"`
+	Enabled bool `yaml:"enabled,omitempty" env:"JUSTIFICATION_ENABLED,overwrite"`
 }
 
 // Validate validates the Justification.
 func (justification *Justification) Validate() error {
-	if justification.Enabled && justification.JVSPublicKeysEndpoint == "" {
-		return fmt.Errorf("jvs_public_keys_endpoint must be specified when justification is enabled")
+	if justification.Enabled {
+		if justification.PublicKeysEndpoint == "" {
+			return fmt.Errorf("public_keys_endpoint must be specified when justification is enabled")
+		}
 	}
 	return nil
 }
--- a/clients/go/apis/v1alpha1/config.go
+++ b/clients/go/apis/v1alpha1/config.go
@@ -65,15 +65,8 @@ type Config struct {
 	// and then continue without impeding the application in any way.
 	LogMode string `yaml:"log_mode,omitempty" env:"LOG_MODE,overwrite"`
 
-	// JVSEndpoint is the endpoint where public keys may be retrieved from the JVS.
-	// These will be used to validate JWT tokens that are passed in through the
-	// "justification-token" header.
-	JVSEndpoint string `yaml:"jvs_endpoint,omitempty" env:"JVS_ENDPOINT,overwrite,default=localhost:8080"`
-
-	// RequireJustification enables adding justification information to audit logs. If this is enabled,
-	// all manual calls are expected to pass in a justification in the "justification-token" header.
-	// If omitted, justifications will not be added to logs, even if provided.
-	RequireJustification bool `yaml:"require_justification,omitempty" env:"REQUIRE_JUSTIFICATION,overwrite,default=false"`
+	// Justification specifies the config used to integrate with JVS.
+	Justification *Justification `yaml:"justification,omitempty" env:",noinit"`
 }
 
 // Validate checks if the config is valid.
@@ -108,15 +101,13 @@ func (cfg *Config) Validate() error {
 		}
 	}
 
-	return err
-}
-
-// ValidateSecurityContext checks if the config SecurityContext is valid.
-func (cfg *Config) ValidateSecurityContext() error {
-	if cfg.SecurityContext == nil {
-		return fmt.Errorf("SecurityContext is nil")
+	if cfg.Justification != nil {
+		if serr := cfg.Justification.Validate(); serr != nil {
+			err = multierr.Append(err, serr)
+		}
 	}
-	return cfg.SecurityContext.Validate()
+
+	return err
 }
 
 // SetDefault sets default for the config.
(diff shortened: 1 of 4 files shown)


clients/go/apis/v1alpha1/config.go
```diff
@@ -358,3 +349,24 @@ func (r *AuditRule) SetDefault() {
 		r.LogType = AuditLogRequest_DATA_ACCESS.String()
 	}
 }
+
+// Justification specifies the config used to integrate with JVS.
+type Justification struct {
+	// JVSPublicKeysEndpoint is the endpoint where public keys may be retrieved from the JVS.
+	// These will be used to validate JWT tokens that are passed in through the
+	// "justification-token" header.
+	JVSPublicKeysEndpoint string `yaml:"jvs_public_keys_endpoint,omitempty"`
```
yolocs, line 358: I think JVS here is redundant info. Should we remove it?

clients/go/apis/v1alpha1/config.go
```diff
@@ -358,3 +349,24 @@ func (r *AuditRule) SetDefault() {
...
+	// all manual calls are expected to pass in a justification in the "justification-token" header.
+	// If omitted, justifications will not be added to logs, even if provided.
+	Enabled bool `yaml:"enabled,omitempty"`
+}
+
+// Validate validates the Justification.
+func (justification *Justification) Validate() error {
+	if justification.Enabled && justification.JVSPublicKeysEndpoint == "" {
+		return fmt.Errorf("jvs_public_keys_endpoint must be specified when justification is enabled")
+	}
```
yolocs, line 370: nit. I recommend do the following. If we need to add configs in the future.

```suggestion
	if !justification.Enabled {
		return nil
	}
	if justification.JVSPublicKeysEndpoint == "" {
		return fmt.Errorf("jvs_public_keys_endpoint must be specified when justification is enabled")
	}
```

clients/go/apis/v1alpha1/config_test.go
```diff
@@ -149,180 +152,226 @@ func TestValidate(t *testing.T) {
 		name    string
 		cfg     *Config
 		wantErr string
-	}{{
```
yolocs, line 152: Please try to keep the current bracelets arrangement to avoid causing unnecessary changes appearing.

clients/go/pkg/auditopt/config.go
```diff
@@ -123,8 +123,8 @@ func interceptorFromConfigFile(ctx context.Context, path string, lookuper envcon
 			return fmt.Errorf("SecurityContext must be provided to use interceptor")
 		}
 
-		if cfg.RequireJustification {
-			jvsClient, err := client.NewJVSClient(ctx, &client.JVSConfig{JVSEndpoint: cfg.JVSEndpoint})
+		if cfg.Justification != nil && cfg.Justification.Enabled {
+			jvsClient, err := client.NewJVSClient(ctx, &client.JVSConfig{JVSEndpoint: cfg.Justification.JVSPublicKeysEndpoint})
```
yolocs, line 127: We now should move this to `clientFromConfig` func since it's no longer gRPC specific?

yolocs (changes requested):

clients/go/apis/v1alpha1/config.go
capri-xiyue: done

clients/go/apis/v1alpha1/config.go
capri-xiyue: I changed it to 
```
	if justification.Enabled {
		if justification.PublicKeysEndpoint == "" {
			return fmt.Errorf("public_keys_endpoint must be specified when justification is enabled")
		}
	}
	return nil
```
I feel this one looks more straightforward

clients/go/apis/v1alpha1/config_test.go
capri-xiyue: The format of this files was not standard, therefore I changed it. I didn't do it manaully, I ran the code formatter in IDE. Therefore, I believe no business logic has been changed.

clients/go/pkg/auditopt/config.go
capri-xiyue: I will do the change once I work on the https://github.com/abcxyz/lumberjack/issues/283

clients/go/apis/v1alpha1/config.go
sethvargo, line 358: These seem to be missing env keys

clients/go/apis/v1alpha1/config.go
```diff
@@ -65,15 +65,8 @@ type Config struct {
...
-	// These will be used to validate JWT tokens that are passed in through the
-	// "justification-token" header.
-	JVSEndpoint string `yaml:"jvs_endpoint,omitempty" env:"JVS_ENDPOINT,overwrite,default=localhost:8080"`
-
-	// RequireJustification enables adding justification information to audit logs. If this is enabled,
-	// all manual calls are expected to pass in a justification in the "justification-token" header.
-	// If omitted, justifications will not be added to logs, even if provided.
-	RequireJustification bool `yaml:"require_justification,omitempty" env:"REQUIRE_JUSTIFICATION,overwrite,default=false"`
+	// Justification specifies the config used to integrate with JVS.
+	Justification *Justification `yaml:"justification,omitempty" env:",noinit"`
```
sethvargo, line 69: Why `noinit`? We have the `Enabled` field to control enablement.

clients/go/pkg/auditopt/config.go
yolocs: Ack

clients/go/apis/v1alpha1/config_test.go
yolocs: I don't think standard go formatter will change this. Please adjust your IDE so future PRs won't have such unintended changes. Thanks.

clients/go/apis/v1alpha1/config.go
yolocs: +1

clients/go/apis/v1alpha1/config_test.go
capri-xiyue: For example, you can see the format of this file is different from the format of  file https://github.com/abcxyz/lumberjack/blob/main/clients/go/apis/v1alpha1/auditlogutil_test.go

sethvargo: The linter should have caught this change when it was submitted. It's possible these changes came in _before_ the linter was connected.

yolocs: @capri-xiyue That's not what I meant. Both styles of bracelet are technically acceptable by gofmt. I don't see a need to force converge to one style.

@sethvargo Are you sure linter will report this as an issue?

clients/go/apis/v1alpha1/config.go
capri-xiyue: Do we want to change the `PublicKeyEndpont` via env variables?
(To clarify, `PublicKeyEndpont`  is field of struct [justification] https://github.com/abcxyz/lumberjack/pull/284/files#diff-be7b293b39821f8b46a4da6a72de39a29d1508d34745442b19d0028a6dbe00bcR69)

yolocs: I think my rule of thumb is to allow config setting/override with env var if possible. We only don't support env var for some repeated fields, e.g. rules.

clients/go/apis/v1alpha1/config.go
capri-xiyue: If user doesn't specify the justification, I prefer lumberjack initialize it to nil instead of empty struct. And it looks like this is the convention in the codebase, see https://github.com/abcxyz/lumberjack/blob/main/clients/go/apis/v1alpha1/config.go#L40-L50

clients/go/apis/v1alpha1/config.go
capri-xiyue: done

clients/go/apis/v1alpha1/config_test.go
sethvargo: Nope :)

clients/go/apis/v1alpha1/config.go
sethvargo: Sure, but if you let envconfig initialize the struct, `Enabled` will default to `false`. Then you don't have to check `justification != nil && justification.Enabled`, you can just always check `justification.Enabled` instead.

capri-xiyue: I think it could be useful and safer if we can differentiate among unset(disable justification), user enable the justification, user disable the justification explicitly. In addition,  the codebase already uses `nolint` for other configs like `Backend`, `Condition` and `SecurityContext`. I prefer to keep the code style consistent.

yolocs (approved):

clients/go/apis/v1alpha1/config.go
capri-xiyue: I will merge this PR for now since https://github.com/abcxyz/lumberjack/issues/283 depends on it. I don't have strong preference against nil or empty struct initialization. If we think empty struct is better, I will change all configs including `Backend`, `Condition` and `SecurityContext` to make the code style consistent in another PR.

PR #285, opened 2022-09-01
Update to Go 1.19 and prep support for arm

sethvargo:
This commit updates the require module versions to Go 1.19 so we can take advantage of some of the faster compile times. It also updates the Dockerfiles to support multi-arch builds, if we ever wanted to provide arm64 or other platform-specific containers.


sethvargo:
I don't know enough about Java ARM support, but this puts all the pieces in place to build Go ARM images if there were customer demand. `buildx` is also generally faster than `docker build`, so there's that benefit too. @yolocs

yolocs (approved):

Issue #257, opened 2022-07-20
Consider adding justification modes


raserva:
I think this is fine for now, but we should consider an enhancement in the future of making this an enum like `JustificationMode`. I can see customer's wanting any of the following behavior:

- Do nothing with justifications
- If a justification is provided, check the signature and add the result as another field (e.g. `justification_verified: true` or `false`)
- If a justification is provided, check the signature and reject the request if it's invalid. If a justification is not provided, accept the request.
- Always require justification and reject requests without justification or requests where justification validation fails

_Originally posted by @sethvargo in https://github.com/abcxyz/lumberjack/pull/254#discussion_r925905502_

yolocs:
I'm having a slightly different list:

- Justification not required (the flag we already have) in which case the code shouldn't even care about justification
- Require justification
  - check the signature; if valid, add the justification to the audit log
  - else if breakglass, add the breakglass justification to the audit log and **continue**
  - else if not valid, reject the request

I think the only difference is "If a justification is not provided, accept the request". I wonder what use case is that?

sethvargo:
I think there's a "dry run" scenario where we want to know whether justification _would_ have succeeded, but we don't want to reject requests. This would be very helpful when rolling out the new system.

yolocs:
We can already turn on / off justification with a sub-block in the configuration.
In the justification config, we can already configure whether allow breakglass.

I'm debating whether we need a "dry run" flag to indicate whether to enforce justification. It seems beneficial per Seth's comment.

When in dryrun, we allow no justification or invalid jvs token. We just log a warning. Otherwise, we demand justification must be present (breakglass or not is a separate knob).

anmluo:
Chen, what will be the use case of we do not need to enforce justification (except breakglass use case)? Based on policy, an Alphabet employee must have a valid business reason when accessing user data. Also, who can turn on/off justification or dry-run mode, and how can we prevent a user from turning justification off by mistake?

yolocs:
After some thought, I think the "dry run" mode can be covered in the "best effort" log mode. When enabled, any error during audit logging will be (std) logged but we won't return the error. The log mode has a global default and can be overridden on each log.

If there is no objection, then I will close this issue.

anmluo:
No objection.

PR #288, opened 2022-09-06
feat: change justification processor logic to make it a general log processor

capri-xiyue:
1. golang part of #283 
2. removed unused function `mustConfigFromFile`


capri-xiyue:
integration test failed because of it can't find logs when testing with GRPC. I'm investigating it.

capri-xiyue:
> integration test failed because of missing logs when testing with GRPC. I'm investigating the failure.

fixed

clients/go/pkg/audit/interceptor.go
```diff
@@ -138,6 +126,15 @@ func (i *Interceptor) UnaryInterceptor(ctx context.Context, req interface{}, inf
...
 	}
+	// look for justification info and set justification token.
+	if i.getJVSToken(ctx) != "" {
+		if logReq.Context == nil {
+			logReq.Context = &structpb.Struct{
+				Fields: map[string]*structpb.Value{},
+			}
+		}
+		logReq.Context.Fields[justification.TokenHeaderKey] = structpb.NewStringValue(i.getJVSToken(ctx))
+	}
```
yolocs, line 137: It seems you can wrap the whole block into a helper func? Also, since the `logReq` is created right above, we can just initialize the `logReq.Context` there.

clients/go/pkg/audit/interceptor.go
```diff
@@ -230,6 +223,16 @@ func (i *Interceptor) StreamInterceptor(srv interface{}, ss grpc.ServerStream, i
...
 
+	// look for justification info and set justification token.
+	if i.getJVSToken(ctx) != "" {
+		if logReq.Context == nil {
+			logReq.Context = &structpb.Struct{
+				Fields: map[string]*structpb.Value{},
+			}
+		}
+		logReq.Context.Fields[justification.TokenHeaderKey] = structpb.NewStringValue(i.getJVSToken(ctx))
+	}
```
yolocs, line 234: Similar comment to the one above.

clients/go/pkg/auditopt/config.go
```diff
@@ -46,40 +46,17 @@ import (
...
-		if err != nil {
-			return err
-		}
-		cfg, err := loadConfig(fc, lookuper)
-		if err != nil {
-			return err
-		}
-		return clientFromConfig(c, cfg)
-	}
-}
```
yolocs, line 70: Why remove?

clients/go/pkg/auditopt/config.go
```diff
@@ -269,6 +246,15 @@ func labelsFromConfig(cfg *api.Config) audit.Option {
 	return audit.WithMutator(&lp)
 }
 
+func justificationFromConfig(ctx context.Context, cfg *api.Config) (audit.Option, error) {
+	jvsClient, err := client.NewJVSClient(ctx, &client.JVSConfig{JVSEndpoint: cfg.Justification.PublicKeysEndpoint})
+	if err != nil {
+		return nil, err
+	}
+	lp := justification.NewProcessor(jvsClient)
+	return audit.WithMutator(lp), nil
```
yolocs, line 255: nit: for such local var, it's ok to use one char var name

```suggestion
	p := justification.NewProcessor(jvsClient)
	return audit.WithMutator(p), nil
```

clients/go/pkg/justification/processor.go
```diff
@@ -51,12 +54,15 @@ func NewProcessor(v Validator) *Processor {
 
 // Process populates the given audit log request with the justification info from the given token.
 // If the token is empty, this function does nothing.
-func (p *Processor) Process(jvsToken string, logReq *api.AuditLogRequest) error {
-	if jvsToken == "" {
+func (p *Processor) Process(ctx context.Context, logReq *api.AuditLogRequest) error {
+	jvsToken, ok := logReq.Context.GetFields()[TokenHeaderKey]
+	jvsTokenStr := jvsToken.GetStringValue()
+	if !ok || jvsTokenStr == "" {
```
yolocs, line 60: Potential bug: if ok is `false`, `jvsToken` could be `nil`. Calling `jvsToken.GetStringValue()` would panic in that case.

`ok` should be checked before calling `jvsToken.GetStringValue()`

clients/go/pkg/justification/processor.go
```diff
@@ -51,12 +54,15 @@ func NewProcessor(v Validator) *Processor {
...
+		log.Println("don't find justification token in field 'context' of AuditLogRequest")
```
yolocs, line 61: No, please use the logger from context if needed. See example: https://github.com/abcxyz/lumberjack/blob/d33b0444e158f6add08e7cd71704aa6cf5719ef3/clients/go/pkg/audit/client.go#L139

Add a TODO with https://github.com/abcxyz/lumberjack/issues/257. We will want to enforce the token later on.

clients/go/test/grpc-app/main.go
```diff
@@ -65,22 +62,25 @@ func realMain() (outErr error) {
...
 	defer shutdown()
 
+	// set Justification config via environment variables.
+	// can't set it via yaml file because we don't know pubKeyEndpoint when we write yaml file.
+	if err := os.Setenv("AUDIT_CLIENT_JUSTIFICATION_PUBLIC_KEYS_ENDPOINT", pubKeyEndpoint); err != nil {
+		return err
+	}
+	if err := os.Setenv("AUDIT_CLIENT_JUSTIFICATION_ENABLED", "true"); err != nil {
+		return err
+	}
```
yolocs, line 72: Nit: but we can set this value in config file, right?

yolocs (changes requested):

clients/go/test/grpc-app/main.go
capri-xiyue: `enabled` can be set in the config file, but do we want to separate the justification setting of `endpoint` and `enabled`? I feel it is a little bit confusing.

clients/go/pkg/auditopt/config.go
capri-xiyue: I can't find the usage of `MustFromConfigFile` in the codebase.

clients/go/pkg/audit/interceptor.go
capri-xiyue: Do we want to  initialize the `logReq.Context`  to an empty struct instead of keeping it nil when there is no jvs token avaialable?

capri-xiyue: helper func added.

clients/go/pkg/audit/interceptor.go
capri-xiyue: similar reply to the one above

clients/go/pkg/justification/processor.go
capri-xiyue: Do you mean the jvs token won't be optional in the future? We will always be able to find the jvs token in the context of grpc requests?

capri-xiyue: changed to zlogger

clients/go/pkg/auditopt/config.go
```diff
@@ -269,6 +246,15 @@ func labelsFromConfig(cfg *api.Config) audit.Option {
 	return audit.WithMutator(&lp)
 }
 
+func justificationFromConfig(ctx context.Context, cfg *api.Config) (audit.Option, error) {
```
sethvargo, line 249: 1. I don't like the idea of creating a new justification client each call.
2. This client is never closed, which could leak fd over time.
3. The threading of `ctx` from parsing the config seems incorrect here.

yolocs: RE 1: This function is supposed to be called once (e.g. in main.go) when creating an audit client. E.g. https://github.com/abcxyz/lumberjack/blob/main/clients/go/test/shell/main.go#L123. In addition, it's possible to not use `auditopt` package to load options from a config file, in which case people can provide a reusable justification client as an option.

RE 2: Not sure I follow. The justification client doesn't have a "close" func. The audit client should be stopped by the caller.

RE 3: How come? The `ctx` is not per-request context.

sethvargo: 1. Okay - that wasn't immediately clear from the PR. Maybe we could add some code comments to clarify the ctx scope.

2. I thought this was a grpc client library (which needs to be closed). I see it's our client library now which does not require closing.

3. See 1

yolocs: Thanks for the callout!

We have some existing doc explaining the usage: https://github.com/abcxyz/lumberjack/blob/main/docs/clients.md#go
But I can see room for improvements and package level examples as godoc. Created an issue: https://github.com/abcxyz/lumberjack/issues/289

clients/go/pkg/justification/processor.go
```diff
@@ -28,6 +30,7 @@ import (
 )
 
 const (
+	TokenHeaderKey              = "justification-token"
 	justificationLogMetadataKey = "justification"
```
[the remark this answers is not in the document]
capri-xiyue: I changed both of them to public even though currently the second one is not used in other packages.

clients/go/test/grpc-app/main.go
yolocs: I think we can set a fake endpoint in the config file and still override it here.

clients/go/pkg/audit/interceptor.go
```diff
@@ -266,28 +252,24 @@ func (i *Interceptor) StreamInterceptor(srv interface{}, ss grpc.ServerStream, i
...
-	// If there is no justification processor, we don't handle justification, and just log
-	// without the justification.
-	if i.justificationProcessor == nil {
-		return nil
-	}
-
-	// Justification processor is present, look for justification info
-	// TODO(#257): If we add justification mode, we need to honor that.
+func fillJVSToken(ctx context.Context, logReq *api.AuditLogRequest) {
+	// look for justification info and set justification token.
```
yolocs, line 256: nit
```suggestion
	// Look for justification info and set justification token.
```

clients/go/pkg/auditopt/config.go
capri-xiyue: `Okay - that wasn't immediately clear from the PR. Maybe we could add some code comments to clarify the ctx scope`, similar usage is also used when we create a logging client https://github.com/abcxyz/lumberjack/blob/f03fda0c297d43a15ce2d367c44b7b1a7c233dfe/clients/go/pkg/auditopt/config.go#L250, instead of adding comments, it will be better to add docs in #289

clients/go/pkg/auditopt/config.go
```diff
@@ -269,6 +246,15 @@ func labelsFromConfig(cfg *api.Config) audit.Option {
...
+	jvsClient, err := client.NewJVSClient(ctx, &client.JVSConfig{JVSEndpoint: cfg.Justification.PublicKeysEndpoint})
```
capri-xiyue, line 250: I found logging client uses `context.TODO()` https://github.com/abcxyz/lumberjack/blob/f03fda0c297d43a15ce2d367c44b7b1a7c233dfe/clients/go/pkg/auditopt/config.go#L250.
I'm wondering do we need to pass the context from the main function? Can we just use `context.TODO()` when we create the jvs client?

yolocs (changes requested):
Some nits

clients/go/test/grpc-app/main.go
capri-xiyue: done

clients/go/pkg/auditopt/config.go
```diff
@@ -246,8 +246,8 @@ func labelsFromConfig(cfg *api.Config) audit.Option {
 	return audit.WithMutator(&lp)
 }
 
-func justificationFromConfig(ctx context.Context, cfg *api.Config) (audit.Option, error) {
-	jvsClient, err := client.NewJVSClient(ctx, &client.JVSConfig{JVSEndpoint: cfg.Justification.PublicKeysEndpoint})
```
yolocs: No, we shouldn't change this.

clients/go/pkg/auditopt/config.go
yolocs: I think passing the context is the right approach. When the context is canceled in main, the jvs client will be shutdown.

yolocs (approved):

PR #291, opened 2022-09-09
feat: refactor JustificationProcessor(Java)

sqin2019:
to address [issue 283](https://github.com/abcxyz/lumberjack/issues/283) java side refactoring

### DIFF
--- a/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
+++ b/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
@@ -119,7 +119,7 @@ public class AuditLoggingServerInterceptor<ReqT extends Message> implements Serv
       next.startCall(call, headers);
     }
 
-    final Struct auditLogReqCtx = getAuditLogRequestContext(headers);
+    final Struct auditLogRequestContext = getAuditLogRequestContext(headers);
 
     LogEntryOperation logEntryOperation =
         LogEntryOperation.newBuilder()
@@ -146,7 +146,7 @@ public class AuditLoggingServerInterceptor<ReqT extends Message> implements Serv
 
                 auditLog(
                     selector,
-                    auditLogReqCtx,
+                    auditLogRequestContext,
                     unloggedRequest,
                     message,
                     logBuilderFinal,
@@ -174,7 +174,7 @@ public class AuditLoggingServerInterceptor<ReqT extends Message> implements Serv
           if (unloggedRequest != null) {
             auditLog(
                 selector,
-                auditLogReqCtx,
+                auditLogRequestContext,
                 unloggedRequest,
                 null,
                 logBuilderFinal,
@@ -197,7 +197,12 @@ public class AuditLoggingServerInterceptor<ReqT extends Message> implements Serv
           log.info("Exception occurred, audit logging it: {}", e.getMessage());
           ReqT unloggedRequest = unloggedRequests.pollFirst(); // try to get the last request
           logError(
-              selector, auditLogReqCtx, unloggedRequest, e, logBuilderFinal, logEntryOperation);
+              selector,
+              auditLogRequestContext,
+              unloggedRequest,
+              e,
+              logBuilderFinal,
+              logEntryOperation);
           throw e;
         }
       }
@@ -206,7 +211,7 @@ public class AuditLoggingServerInterceptor<ReqT extends Message> implements Serv
 
   <ReqT, RespT> void auditLog(
       Selector selector,
-      Struct auditLogReqCtx,
+      Struct auditLogRequestContext,
       ReqT request,
       RespT response,
       AuditLog.Builder logBuilder,
@@ -226,7 +231,7 @@ public class AuditLoggingServerInterceptor<ReqT extends Message> implements Serv
     Instant now = clock.instant();
     builder.setTimestamp(
         Timestamp.newBuilder().setSeconds(now.getEpochSecond()).setNanos(now.getNano()));
-    builder.setContext(auditLogReqCtx);
+    builder.setContext(auditLogRequestContext);
 
     try {
       log.info("Audit logging...");
@@ -242,7 +247,7 @@ public class AuditLoggingServerInterceptor<ReqT extends Message> implements Serv
    */
   <ReqT, RespT> void logError(
       Selector selector,
-      Struct auditLogReqCtx,
+      Struct auditLogRequestContext,
       ReqT request,
       Exception e,
       AuditLog.Builder logBuilder,
@@ -256,7 +261,7 @@ public class AuditLoggingServerInterceptor<ReqT extends Message> implements Serv
     }
     logBuilder.setStatus(
         Status.newBuilder().setCode(code.getNumber()).setMessage(code.name()).build());
-    auditLog(selector, auditLogReqCtx, request, null, logBuilder, logEntryOperation);
+    auditLog(selector, auditLogRequestContext, request, null, logBuilder, logEntryOperation);
   }
 
   /**
@@ -319,7 +324,7 @@ public class AuditLoggingServerInterceptor<ReqT extends Message> implements Serv
    */
   Struct getAuditLogRequestContext(Metadata headers) {
     String jvsToken = headers.get(JUSTIFICATION_METADATA_KEY);
-    if (jvsToken != null && jvsToken.length() > 0) {
+    if (jvsToken != null && !jvsToken.isEmpty()) {
       return Struct.newBuilder()
           .putFields(
               JUSTIFICATION_TOKEN_HEADER_KEY, Value.newBuilder().setStringValue(jvsToken).build())
--- a/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
+++ b/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
@@ -63,10 +63,11 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public class AuditLoggingServerInterceptor<ReqT extends Message> implements ServerInterceptor {
   public static final String JUSTIFICATION_TOKEN_HEADER_KEY = "justification-token";
-  private static final Metadata.Key<String> JUSTIFICATION_METADATA_KEY =
-      Metadata.Key.of(JUSTIFICATION_TOKEN_HEADER_KEY, Metadata.ASCII_STRING_MARSHALLER);
   public static final Context.Key<AuditLog.Builder> AUDIT_LOG_CTX_KEY = Context.key("audit-log");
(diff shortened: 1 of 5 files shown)


clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
```diff
@@ -207,6 +206,7 @@ public void onHalfClose() {
 
   <ReqT, RespT> void auditLog(
       Selector selector,
+      Struct auditLogReqCtx,
```
yolocs, line 209: Instead of adding a new parameter, can we just set the `auditLogReqCtx` in the `logBuilder`?

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/LoggingClientBuilder.java
```diff
@@ -74,14 +79,19 @@ private <T> T getProcessor(Class<T> clazz) {
...
+   * {@link JustificationProcessor}. }
    */
   public LoggingClientBuilder withDefaultProcessors() {
-    return withValidationProcessor()
+    withValidationProcessor()
         .withFilteringProcessor()
         .withRuntimeInfoProcessor()
         .withDefaultLogBackends()
         .withLabelProcessor();
+    if (auditLoggingConfiguration.isJustificationRequired()) {
```
yolocs, line 91: Should be the `if` logic in the `withJustificationProcessor` method? Similar to `withDefaultLogBackends`

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
```diff
@@ -30,30 +32,43 @@
...
     AuditLogRequest.Builder auditLogRequestBuilder = auditLogRequest.toBuilder();
     AuditLog.Builder auditLogBuilder = auditLogRequest.getPayload().toBuilder();
-    auditLogBuilder = this.auditLogBuilderWithJustification(jvsToken, auditLogBuilder);
+
+    Value jvsToken =
+        auditLogRequest
+            .getContext()
+            .getFieldsMap()
+            .get(AuditLoggingServerInterceptor.JUSTIFICATION_TOKEN_HEADER_KEY);
+    if (jvsToken == null || "".equals(jvsToken.getStringValue())) {
```
yolocs, line 65: Could you add `TODO(#257): JVS token might be required in the future`

yolocs: https://github.com/abcxyz/lumberjack/issues/257

yolocs (changes requested):
Congrats on the first PR in lumberjack! Only some nits

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
sqin2019: Sure, as metadata?

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/LoggingClientBuilder.java
sqin2019: I was hesitating between what you said and what I did here, my reason was that withJustificationProcessor sounds more like it will add justification processor, but I am fine with either.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
sqin2019: Sure, will add.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
```diff
@@ -118,22 +119,7 @@ public <ReqT, RespT> Listener<ReqT> interceptCall(
...
-        if (ConfigUtils.shouldFailClose(auditLoggingConfiguration.getLogMode())) {
-          throw new IllegalStateException("Unable to determine justification.", e);
-        } else {
-          log.error(
-              "Justification was unable to be determined, continuing without audit logging.", e);
-          next.startCall(call, headers);
-        }
-      }
-    }
+    final Struct auditLogReqCtx = getAuditLogRequestContext(headers);
```
myurtoglu, line 122: In Java (in contrast to Go), we should avoid abbreviating names and spell them out as much as we can, so this one should be `final Struct auditLogRequestContext = ...`

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
```diff
@@ -313,14 +315,17 @@ Optional<Selector> getRelevantSelector(String methodIdentifier) {
...
    */
-  AuditLog.Builder setJustification(Metadata headers, AuditLog.Builder logBuilder)
-      throws LogProcessingException {
-    Metadata.Key<String> metadataKey =
-        Metadata.Key.of("justification-token", Metadata.ASCII_STRING_MARSHALLER);
-    String jvsToken = headers.get(metadataKey);
-    return justificationProcessor.auditLogBuilderWithJustification(jvsToken, logBuilder);
+  Struct getAuditLogRequestContext(Metadata headers) {
+    String jvsToken = headers.get(JUSTIFICATION_METADATA_KEY);
+    if (jvsToken != null && jvsToken.length() > 0) {
```
myurtoglu, line 322: For checking if string is not empty we should use `!jvsToken.isEmpty()` or `!jvsToken.isBlank()` rather than the length check.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
yolocs: As context: https://github.com/abcxyz/lumberjack/blob/main/protos/v1alpha1/audit_log_request.proto#L89
Essentially what you did in the method: `builder.setContext(auditLogReqCtx);`

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
myurtoglu, line 65: Similar to above comment, how about `jvsToken.getStringValue().isEmpty()` instead?

clients/java-logger/library/src/test/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessorTests.java
```diff
@@ -102,15 +106,25 @@ public void processShouldThrowExceptionWhenFailingToValidateToken() throws Excep
...
+  }
+
+  @Test
+  public void processLogReqWithNullJVSToken() throws Exception {
+    verifyNoOp(auditLogRequest);
+  }
+
+  @Test
+  public void processLogReqWithEmptyJVSToken() throws Exception {
+    verifyNoOp(getAuditLogRequest(""));
```
myurtoglu, line 127: Could we name the helper to something like `getAuditLogRequestWithJvsToken("")` to make it obvious that the empty string we are passing is the JVS token?

myurtoglu (changes requested):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
sqin2019: hum...I don't see the context field in logBuilder (AuditLog.Builder). If to set it in the AuditLogRequest, since this request is initialized in the auditLog method, have to pass it in either as a parameter or in the logBuilder (AuditLog.Builder) as metadata I assume.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
sqin2019: Ack, will change.

clients/java-logger/library/src/test/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessorTests.java
sqin2019: Yes, we could, will change.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
sqin2019: discussed offline, agreed to keep it as it is.

capri-xiyue (approved):

yolocs (approved):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/LoggingClientBuilder.java
```diff
@@ -74,14 +79,16 @@ private <T> T getProcessor(Class<T> clazz) {
...
+   * {@link JustificationProcessor}. }
    */
   public LoggingClientBuilder withDefaultProcessors() {
     return withValidationProcessor()
         .withFilteringProcessor()
         .withRuntimeInfoProcessor()
         .withDefaultLogBackends()
-        .withLabelProcessor();
+        .withLabelProcessor()
+        .withJustificationProcessor();
```
myurtoglu, line 91: Nit: I would rename this to something like `.withJustificationProcessorIfRequired()` in order to convey the fact that if the configuration doesn't require it, it won't be added to mutators even if this is called.

myurtoglu (approved):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/LoggingClientBuilder.java
sqin2019: Ok, will rename it in the following CR for https://github.com/abcxyz/lumberjack/issues/280 to `.withJustificationProcessorIfEnabled` where the env var is also changed (from REQUIRED to ENBLED).

myurtoglu: SG, thanks.

PR #292, opened 2022-09-14
fix: refactor justification config in java client

sqin2019:
To address [issue 280](https://github.com/abcxyz/lumberjack/issues/280)


clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/AuditLoggingConfiguration.java
```diff
@@ -46,11 +46,8 @@ public class AuditLoggingConfiguration {
   @JsonProperty("security_context")
   private SecurityContext securityContext;
 
-  @JsonProperty(value = "jvs_endpoint")
-  private String jvsEndpoint = "localhost:8080";
-
-  @JsonProperty(value = "require_justification")
-  private boolean justificationRequired = false;
+  @JsonProperty("justification")
+  private Justification justification = new Justification();
```
capri-xiyue, line 50: Not quite familiar with Java. Do we need new Justification() here? Can we just have nil value as the default value?

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/Justification.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.Data;
+
+@Data
+public class Justification {
+  private static final String JUSTIFICATION_PUBLIC_KEYS_ENDPOINT_ENV_KEY =
+      "AUDIT_CLIENT_JUSTIFICATION_PUBLIC_KEYS_ENDPOINT";
+  private static final String JUSTIFICATION_ENABLED_ENV_KEY = "AUDIT_CLIENT_JUSTIFICATION_ENABLED";
+
+  @JsonProperty("public_keys_endpoint")
+  private String publicKeysEndpoint = "localhost:8080";
```
capri-xiyue, line 29: To be consistent with the logic in Golang part(https://github.com/abcxyz/lumberjack/pull/284 fyi), we no longer use "localhost:8080" as the default value. There is no default value for the public_keys_endpoint

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/AuditLoggingConfiguration.java
sqin2019: This will create a default justification object to keep the default values as they were. It is indifferent in terms of outcome comparing to set it to null.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/Justification.java
```diff
@@ -0,0 +1,45 @@
...
+  @JsonProperty("enabled")
+  private boolean enabled;
+
+  public String getPublicKeysEndpoint() {
+    return System.getenv()
+        .getOrDefault(JUSTIFICATION_PUBLIC_KEYS_ENDPOINT_ENV_KEY, publicKeysEndpoint);
+  }
+
+  public boolean isEnabled() {
+    if (System.getenv().containsKey(JUSTIFICATION_ENABLED_ENV_KEY)) {
```
capri-xiyue, line 40: Can you created a shared util for it? FYI: https://github.com/abcxyz/lumberjack/blob/main/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/Filters.java#L41-L45

sqin2019: Sure, will add.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/AuditLoggingConfiguration.java
myurtoglu: I'm guessing that if this were to default to null, `auditLoggingConfiguration.getJustification().isEnabled()` would throw null pointer exception, so this seems good to me.

capri-xiyue:
In java, we also need the validation similar to Golang: https://github.com/abcxyz/lumberjack/blob/main/clients/go/apis/v1alpha1/config.go#L104-L108

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/AuditLoggingConfiguration.java
capri-xiyue: Looks like for `SecurityContext`, `ConditionConfig` they use the null as the default value. And here is the similar discussion in Golang: https://github.com/abcxyz/lumberjack/pull/284#discussion_r960092449

scripts/format_java.sh
```diff
@@ -26,4 +26,4 @@ then
   chmod a+x .local/$JAR_NAME
 fi
 
-java -jar .local/$JAR_NAME -i **/*
```
myurtoglu, line 29: Was the original **/* not matching every file?

sqin2019: Did it work for you? when I tried to run bash scripts/formate_java.sh, it would not check files recursively, it stopped at the second level of the directory. Should I use a different command?

sqin2019: Here is the link where I find working after apply the change: https://github.com/google/google-java-format/issues/129

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/Justification.java
sqin2019: Any particular reason that we should not use this default value?

capri-xiyue: "localhost:8080" doesn't make sense. In production, people won't use "localhost:8080" to connect to jvs

scripts/format_java.sh
myurtoglu: I hadn't used this script before (assumed it was working), and looks like it's not working, so thanks for fixing it, sg.

myurtoglu (approved):

capri-xiyue:
Can you follow https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention to update the title of this PR?

.gitignore
```diff
@@ -61,3 +61,6 @@ buildNumber.properties
 
 # Go workspace file
 go.work
+
+### Visual Studio Code IDEA ###
+.vscode
```
yolocs, line 66: FYI, it's possible to have vscode setting outside of your code repo.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/Justification.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2021 Lumberjack authors (see AUTHORS file)
```
yolocs, line 2:
```suggestion
 * Copyright 2022 Lumberjack authors (see AUTHORS file)
```

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/Justification.java
yolocs: +1 We no longer set default for this value in Go https://github.com/abcxyz/lumberjack/blob/main/clients/go/apis/v1alpha1/config.go#L358

yolocs:
> Can you follow [kubeflow/pipelines@`master`/CONTRIBUTING.md#pull-request-title-convention](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md?rgh-link-date=2022-09-14T19%3A19%3A24Z#pull-request-title-convention) to update the title of this PR?

This is canonical version: https://www.conventionalcommits.org/en/v1.0.0/#summary

yolocs (changes requested):
Title name should be something like `fix: refactor justification config in java client`

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/AuditLoggingConfiguration.java
sqin2019: Yes, since Go client set it to null, will make the change for consistancy.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/Justification.java
sqin2019: Ack, thanks!

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/Justification.java
sqin2019: nice catch! fixed.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/AuditLoggingConfiguration.java
capri-xiyue: At first, I didn't catch the point that Mete mentioned that `auditLoggingConfiguration.getJustification().isEnabled() would throw null pointer exception`. In go client, when we use `justification`, to avoid the null pointer exception, we will check whether `justisfication` is nil or not.  https://github.com/abcxyz/lumberjack/blob/563a9dd75e25d3645e6bebecc506d7770679f17c/clients/go/pkg/auditopt/config.go#L173.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/Justification.java
```diff
@@ -26,20 +27,24 @@ public class Justification {
...
-  private String publicKeysEndpoint = "localhost:8080";
+  private String publicKeysEndpoint;
 
   @JsonProperty("enabled")
   private boolean enabled;
 
   public String getPublicKeysEndpoint() {
-    return System.getenv()
-        .getOrDefault(JUSTIFICATION_PUBLIC_KEYS_ENDPOINT_ENV_KEY, publicKeysEndpoint);
+    return ConfigUtils.getEnvOrDefault(
```
yolocs, line 36: For future, let's separate such refactoring to be its own PR. They are not essential to the change at hand and cause distraction for reviewers.

yolocs (approved):

capri-xiyue (approved):

PR #293, opened 2022-09-14
feat: Populate standard audit log field for justification

capri-xiyue:
fixed golang part of #266

### DIFF
--- a/clients/go/pkg/justification/processor.go
+++ b/clients/go/pkg/justification/processor.go
@@ -95,19 +95,20 @@ func (p *Processor) Process(ctx context.Context, logReq *api.AuditLogRequest) er
 
 	justs, ok := (*tok).Get("justs")
 	if !ok {
-		return fmt.Errorf("can't find 'justs' in claims, denying")
+		logger.Warn("can't find 'justs' in claims")
+	} else {
+		justsBytes, err := json.Marshal(justs)
+		if err != nil {
+			return fmt.Errorf("failed to marshal 'justs'")
+		}
+		if logReq.Payload.RequestMetadata == nil {
+			logReq.Payload.RequestMetadata = &audit.RequestMetadata{}
+		}
+		if logReq.Payload.RequestMetadata.RequestAttributes == nil {
+			logReq.Payload.RequestMetadata.RequestAttributes = &attribute_context.AttributeContext_Request{}
+		}
+		logReq.Payload.RequestMetadata.RequestAttributes.Reason = string(justsBytes)
 	}
-	justsBytes, err := json.Marshal(justs)
-	if err != nil {
-		return fmt.Errorf("failed to marshal 'justs', denying")
-	}
-	if logReq.Payload.RequestMetadata == nil {
-		logReq.Payload.RequestMetadata = &audit.RequestMetadata{}
-	}
-	if logReq.Payload.RequestMetadata.RequestAttributes == nil {
-		logReq.Payload.RequestMetadata.RequestAttributes = &attribute_context.AttributeContext_Request{}
-	}
-	logReq.Payload.RequestMetadata.RequestAttributes.Reason = string(justsBytes)
 
 	return nil
 }
--- a/clients/go/pkg/justification/processor.go
+++ b/clients/go/pkg/justification/processor.go
@@ -23,6 +23,8 @@ import (
 
 	zlogger "github.com/abcxyz/pkg/logging"
 	"github.com/lestrrat-go/jwx/v2/jwt"
+	"google.golang.org/genproto/googleapis/cloud/audit"
+	"google.golang.org/genproto/googleapis/rpc/context/attribute_context"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/types/known/structpb"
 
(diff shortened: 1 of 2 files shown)


clients/go/pkg/justification/processor.go
```diff
@@ -90,6 +92,22 @@ func (p *Processor) Process(ctx context.Context, logReq *api.AuditLogRequest) er
 	}
 
 	logReq.Payload.Metadata.Fields[LogMetadataKey] = structpb.NewStructValue(&tokStruct)
-	// TODO(#266): Also populate RequestAttribute.Reason.
+
+	justs, ok := (*tok).Get("justs")
+	if !ok {
+		return fmt.Errorf("can't find 'justs' in claims, denying")
+	}
```
yolocs, line 99: Let's log warning instead.

clients/go/pkg/justification/processor_test.go
```diff
@@ -78,6 +79,11 @@ func TestProcess(t *testing.T) {
 						}),
 					},
 				},
+				RequestMetadata: &audit.RequestMetadata{
+					RequestAttributes: &attribute_context.AttributeContext_Request{
+						Reason: "[{\"category\":\"explanation\",\"value\":\"need-access\"}]",
```
yolocs, line 84:
```suggestion
						Reason: `[{"category":"explanation","value":"need-access"}]`,
```

yolocs (commented):
Unit test seems failing

clients/go/pkg/justification/processor.go
```diff
@@ -90,6 +92,22 @@ func (p *Processor) Process(ctx context.Context, logReq *api.AuditLogRequest) er
 	}
 
 	logReq.Payload.Metadata.Fields[LogMetadataKey] = structpb.NewStructValue(&tokStruct)
-	// TODO(#266): Also populate RequestAttribute.Reason.
+
+	justs, ok := (*tok).Get("justs")
```
yolocs, line 96: Do you want what's the actual type of `justs` (the returned type is `interface{}`) - the returned value.

capri-xiyue: Maybe I didn't get the context. We need the encoded json of `justs` and `json.Marshal` accepts `interface{}`. Why do we care about the actual type of `justs`?

capri-xiyue:
> Unit test seems failing

fixed

clients/go/pkg/justification/processor.go
yolocs: What if it's already a json string? Also out of curiosity :)

clients/go/pkg/justification/processor.go
```diff
@@ -90,6 +92,23 @@ func (p *Processor) Process(ctx context.Context, logReq *api.AuditLogRequest) er
...
+	if !ok {
+		logger.Warn("can't find 'justs' in claims")
```
yolocs, line 98: nit: try to avoid unnecessary nesting.
```suggestion
	if !ok {
		logger.Warn("can't find 'justs' in claims, skip logging RequestMetadata.RequestAttributes.Reason")
		return nil
	}
```

yolocs (approved):

clients/go/pkg/justification/processor.go
capri-xiyue: To do this, we have to make sure the `reason` field is always that  last field we handle. In the future, if we need to populate other fields, it is easy to introduce a bug that we exit early.

clients/go/pkg/justification/processor.go
capri-xiyue:
```
	bytes, err := json.Marshal(`[{"category":"explanation","value":"need-access"}]`)
			if err != nil {
				return
			}
			fmt.Println(string(bytes))
```
The format is the same

clients/go/pkg/justification/processor.go
yolocs: Fair enought

clients/go/pkg/justification/processor.go
yolocs: I know the marshaling a string is still the same string. But I'm still curious what's the original type from the token. This is not blocking.

capri-xiyue: Looks like the type is a list of `justification` https://github.com/abcxyz/jvs/blob/5ab0828c78626fee6284de86035be5ebb558565f/pkg/justification/processor.go#L160 and https://github.com/abcxyz/jvs/blob/5ab0828c78626fee6284de86035be5ebb558565f/client-lib/go/client/jvs_client_test.go#L238

clients/go/pkg/justification/processor.go
```diff
@@ -90,6 +92,23 @@ func (p *Processor) Process(ctx context.Context, logReq *api.AuditLogRequest) er
 	}
 
 	logReq.Payload.Metadata.Fields[LogMetadataKey] = structpb.NewStructValue(&tokStruct)
-	// TODO(#266): Also populate RequestAttribute.Reason.
+
+	justs, ok := (*tok).Get("justs")
```
sethvargo, line 96: I'm not sure I understand why we need to dereference this? Same with the call to json.Marshal above.

capri-xiyue: The return of function `ValidateJWT` is a pointer of token
 pointer https://github.com/abcxyz/lumberjack/blob/29ad85da8bbc00812ba4023576d9022c3f53b0a0/clients/go/pkg/justification/processor.go#L42? Do you mean we should change the interface of `ValidateJWT` here?

yolocs: I think the question is why we need to dereference the pointer. Why can't we use it directly. TBH I didn't even pay attention to that :P

capri-xiyue: `Get` is a method of interface `Token` instead of interface of Token Pointer. To call `Get`, I need to deference to the pointer.

sethvargo: You can call the function regardless of the receiver: https://go.dev/play/p/dyDG1YNBU0U

We should also always aim to define functions on the pointer receivers.

If `jwt.Token` is an interface (looks like it is), we should update our interface to return the interface, not the pointer to the interface.

sethvargo: I think I see what's going on: https://github.com/abcxyz/jvs/pull/117

sethvargo: And https://github.com/abcxyz/jvs/issues/116

PR #294, opened 2022-09-15
Don't use pointers to interfaces

sethvargo:
Depends on https://github.com/abcxyz/jvs/pull/117 first


myurtoglu (approved):

yolocs:
Thanks!

sethvargo:
Tests seem to be failing because of https://github.com/abcxyz/lumberjack/pull/295

yolocs:
> Tests seem to be failing because of #295

Thanks for the fix in the other PR. 
But the error seems rather accidental.

> Err while reading fibonacci stream: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (1702000233 vs. 4194304)

Let me retry the test. Meantime let me investigate why the error happened in the first place.

yolocs:
I found the root cause of the test failure. We pulled in a new version of JVS in which it changed the logic to use `envconfig` to set config defaults. But in the lumberjack code we didn't load the JVS client config in the right way so JVS client can't be created. The test failure was directly caused by the test app failing to start.

I will send a fix soon.

yolocs:
That said, I expect the test will still fail here :(

capri-xiyue (approved):

Mithu147 (approved):

PR #295, opened 2022-09-15
Fix a NPE bug in grpc test

sethvargo:
These tests will NPE because they call a value on a pointer receiver that is nil if there's an error.


integration/testrunner/grpcrunner/grpc.go
```diff
@@ -158,6 +158,7 @@ func (g *GRPC) runFibonacciCheck(ctx context.Context, tb testing.TB) {
 		}
 		if err != nil {
 			tb.Errorf("Err while reading fibonacci stream: %v", err)
+			continue
```
sethvargo, line 161: Maybe we mean to `return` here instead? But calling `place.Value` on 162 will NPE as-is.

yolocs: Thanks for detecting this! I think we probably want to `return` here to avoid infinite retry.

The error message says:

> Err while reading fibonacci stream: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (1702000233 vs. 4194304)

(I need to investigate why the error happened in the first place)
It seems retrying won't even help in this case.

sethvargo: Switched to break (`return` will stop the test). We could also switch to `Fatalf` if you think that's more appropriate.

capri-xiyue (approved):

PR #304, opened 2022-09-15
feat: populate standard audit log field for justification(java client)

sqin2019:
java client side change for issue #266

### DIFF
--- a/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
+++ b/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
@@ -128,17 +128,16 @@ public class JustificationProcessor implements LogMutator {
       auditLogBuilderCopy.setMetadata(metadataBuilder.build());
 
       Claim justificationsClaim = jwt.getClaim("justs");
+
+      // Continue without populating RequestMetadata.RequestAttributes.Reason if the justifications
+      // claim is null or empty.
       if (justificationsClaim.isNull()) {
-        log.warn(
-            "can't find 'justs' in claims, continue without logging"
-                + " RequestMetadata.RequestAttributes.Reason");
+        log.warn("can't find 'justs' in claims");
         return auditLogBuilderCopy;
       }
       List<Map> justificationsList = justificationsClaim.asList(Map.class);
       if (justificationsList == null || justificationsList.isEmpty()) {
-        log.warn(
-            "justs in claims is null or empty, continue without logging"
-                + " RequestMetadata.RequestAttributes.Reason");
+        log.warn("justs in claims is null or empty");
         return auditLogBuilderCopy;
       }
 
--- a/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
+++ b/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
@@ -21,12 +21,14 @@ import com.abcxyz.lumberjack.auditlogclient.AuditLoggingServerInterceptor;
 import com.abcxyz.lumberjack.auditlogclient.processor.LogProcessor.LogMutator;
 import com.abcxyz.lumberjack.v1alpha1.AuditLogRequest;
 import com.auth0.jwk.JwkException;
+import com.auth0.jwt.exceptions.JWTDecodeException;
 import com.auth0.jwt.interfaces.Claim;
 import com.auth0.jwt.interfaces.DecodedJWT;
 import com.google.api.client.util.Preconditions;
 import com.google.api.client.util.StringUtils;
 import com.google.cloud.audit.AuditLog;
 import com.google.cloud.audit.RequestMetadata;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.Gson;
 import com.google.inject.Inject;
 import com.google.protobuf.InvalidProtocolBufferException;
@@ -45,6 +47,7 @@ import org.apache.commons.codec.binary.Base64;
 @AllArgsConstructor(onConstructor = @__({@Inject}))
 public class JustificationProcessor implements LogMutator {
   private static final String JUSTIFICATION_LOG_METADATA_KEY = "justification";
+  private static final String JUSTIFICATIONS_CLAIM_NAME = "justs";
   private final JvsClient jvs;
 
   /**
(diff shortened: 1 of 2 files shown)


clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
```diff
@@ -121,6 +126,20 @@ public AuditLog.Builder auditLogBuilderWithJustification(
           Value.newBuilder().setStructValue(justificationStruct).build());
       auditLogBuilderCopy.setMetadata(metadataBuilder.build());
 
+      Claim justificationsClaim = jwt.getClaim("justs");
+      if (justificationsClaim.isNull()) {
+        log.warn("can't find 'justs' in claims");
+      } else {
+        RequestMetadata.Builder requestMetadataBuilder =
+            auditLogBuilderCopy.getRequestMetadataBuilder();
```
capri-xiyue, line 134: I saw that line 123 uses `auditLogBuilderCopy.getMetadata().toBuilder()`. Does requestMetadata has the same function `auditLogBuilderCopy.getRequestMetadat().toBuilder()`? If there is, what's the difference?

clients/java-logger/library/src/test/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessorTests.java
```diff
@@ -78,14 +92,31 @@ public void processShouldPopulateJustificationMetadata() throws Exception {
...
     Struct wantMetadata =
         Struct.newBuilder()
             .putFields(
                 "justification", Value.newBuilder().setStructValue(wantJustification).build())
             .build();
+    RequestMetadata wantRequestMetadata =
+        RequestMetadata.newBuilder()
+            .setRequestAttributes(
+                Request.newBuilder()
+                    .setReason("[{\"category\":\"explanation\",\"value\":\"need-access\"}]")
```
capri-xiyue, line 112: Does `[{"category":"explanation","value":"need-access"}]` work?

sqin2019: No, I tried that, not sure why.

capri-xiyue: What's the error message?

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
sqin2019: GetXXBuilder() or getXX().toBuilder() will return a default/empty instance if it is null. The different between the two method is one returns a Builder one returns an object, use getXXBuilder() since it is all we need otherwise will need to convert the object to builder to set values in it.

The logic is what you mentioned, it only adds the reason and does not alter anything else when justs is found and non0null/empty.

clients/java-logger/library/src/test/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessorTests.java
sqin2019:
```
java.lang.Error: Unresolved compilation problems: 
 The method setReason(String) in the type AttributeContext.Request.Builder is not applicable for the arguments (String, String)
 Syntax error on tokens, delete these tokens
```
looks like it still need escape for "" inside the sentence.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
```diff
@@ -121,6 +127,26 @@ public AuditLog.Builder auditLogBuilderWithJustification(
           Value.newBuilder().setStructValue(justificationStruct).build());
       auditLogBuilderCopy.setMetadata(metadataBuilder.build());
 
+      Claim justificationsClaim = jwt.getClaim("justs");
```
capri-xiyue, line 130: Can we add a comment that the logic here is no justs and then continue with other the populate of other fields?
Context: https://github.com/abcxyz/lumberjack/pull/293#discussion_r971292134

sqin2019: I have added a change to return early when justs is not found or null or empty, do you think it is enough?

sqin2019: Ok, I just see your comment there, for the concern that missing populating other new fields, I think unit test will catch that, what do you think?

capri-xiyue: Why the unit tests will catch it?For example, later we when want to populate another field called "C", the developer need to know the logic is "C" filed still need to get populated even though there is no justs found. The developer need to know the logic and then modify the UT to catch that. Our current UT won't be able to catch it.

sqin2019: Right, current UT should only cover existing functionalities, if adding a new field, the developer should add UT to test that is what I meant.

capri-xiyue: For the developers to add UT, they need to know the business logic is to continue when there is no justs found. That's why I asked to add a comment. Currently the code looks unclear whether the logic is to exit or to continue.

sqin2019: Added to avoid future confusions, please check newest commit, thanks!

capri-xiyue: Can we add a Java comments instead of adding it in the logging?

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
capri-xiyue: ack. Thank you for the explanation and confirmation!

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
sqin2019: Ack.

capri-xiyue (approved):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
```diff
@@ -121,6 +127,29 @@ public AuditLog.Builder auditLogBuilderWithJustification(
...
+
+      RequestMetadata.Builder requestMetadataBuilder =
+          auditLogBuilderCopy.getRequestMetadataBuilder();
+      Request.Builder requestAttributesBuilder =
+          requestMetadataBuilder
+              .getRequestAttributesBuilder()
+              .setReason(new Gson().toJson(justificationsList));
+      requestMetadataBuilder =
+          requestMetadataBuilder.setRequestAttributes(requestAttributesBuilder);
+      auditLogBuilderCopy.setRequestMetadata(requestMetadataBuilder);
```
yolocs, line 152: What if there is an existing requestMetadata in the audit log?

yolocs (commented):
cc: @myurtoglu for a quick java readability review

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
capri-xiyue: Context: https://github.com/abcxyz/lumberjack/pull/304#discussion_r972475115

raserva: I think its worth adding a unit test for this.

clients/java-logger/library/src/test/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessorTests.java
```diff
@@ -78,14 +92,31 @@ public void processShouldPopulateJustificationMetadata() throws Exception {
...
+                    .build())
```
raserva, line 113: We're missing negative tests - validate that the request metadata isn't populated on null/empty justs

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
```diff
@@ -121,6 +127,29 @@ public AuditLog.Builder auditLogBuilderWithJustification(
           Value.newBuilder().setStructValue(justificationStruct).build());
       auditLogBuilderCopy.setMetadata(metadataBuilder.build());
 
+      Claim justificationsClaim = jwt.getClaim("justs");
```
raserva, line 130: better to have a constant for this string

sqin2019: ack.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessor.java
sqin2019: I will go back to the common way of get the object instead of builder since multiple people are having doubts on it.

clients/java-logger/library/src/test/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessorTests.java
sqin2019: We have a ut setLogJustificationShouldPopulateJustificationMetadata in which request metadata is skipped when justs is null, I will create seperate ut when list is null or empty, thanks for pointing it out!

clients/java-logger/library/src/test/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessorTests.java
```diff
@@ -78,14 +93,31 @@ public void processShouldPopulateJustificationMetadata() throws Exception {
...
     Struct wantMetadata =
         Struct.newBuilder()
             .putFields(
                 "justification", Value.newBuilder().setStructValue(wantJustification).build())
             .build();
+    RequestMetadata wantRequestMetadata =
+        RequestMetadata.newBuilder()
+            .setRequestAttributes(
+                Request.newBuilder()
+                    .setReason("[{\"category\":\"explanation\",\"value\":\"need-access\"}]")
```
yolocs, line 113: @capri-xiyue Can you confirm on the Go side we're handling it as a list as well?

yolocs (approved):

clients/java-logger/library/src/test/java/com/abcxyz/lumberjack/auditlogclient/processor/JustificationProcessorTests.java
sqin2019: Checked JVS java client, it is also using the [justs claim as a list](https://github.com/abcxyz/jvs/blob/e718d4664467b880991b8e2a400070c2aa93a0b9/client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java#L76). Also tried to convert the justs Claim directly to json using both gson and jackson, neither of them worked because Claim holds the value in a generic way.

raserva (approved):

PR #306, opened 2022-09-23
Update to latest jvs libraries and drop golang-jwt

sethvargo:


yolocs (approved):

clients/go/test/shell/main.go
```diff
@@ -42,44 +45,60 @@ const (
...
+	logger = logger.With("trace_id", traceID)
+	logger.Debugw("found trace id")
+
+	// Parse the JWT. We do not verify the JWT because:
+	//
+	//   - This app is only for testing purposes. It should never be used for anything else.
+	//   - This runs as a Cloud Run service, and Cloud IAM verifies the JWT.
+	//   - Cloud Run actually strips the signature, so there is no signature to verify.
+	idToken := idTokenFrom(r.Header.Get("Authorization"))
+	token, err := jwt.ParseString(idToken, jwt.WithVerify(false))
```
yolocs, line 78: I think it's this line. I think we need to remove the `Bearer ` prefix

PR #310, opened 2022-09-26
chore: update JVS dependency and use cfgloader for JVS config

yolocs:
Part of: https://github.com/abcxyz/lumberjack/issues/299
Will use `cfgloader` for lumberjack config once https://github.com/abcxyz/pkg/pull/38 is merged.


clients/go/pkg/auditopt/config.go
```diff
@@ -260,13 +260,11 @@ func labelsFromConfig(cfg *api.Config) audit.Option {
 }
 
 func justificationFromConfig(ctx context.Context, cfg *api.Config) (audit.Option, error) {
-	// TODO(#299): allow overriding these values via JVS env var.
-	jvsconfig := &client.JVSConfig{
-		JWKSEndpoint: cfg.Justification.PublicKeysEndpoint,
-		Version:      "1",
-		CacheTimeout: 5 * time.Minute,
+	var jvsconfig client.JVSConfig
+	if err := cfgloader.Load(ctx, &jvsconfig, cfgloader.WithEnvPrefix("JVS_")); err != nil {
```
sethvargo, line 264: I thought we were dropping the `JVS_` prefix

sethvargo (approved):
One small nit - approving because I don't want to block on it.

clients/go/pkg/auditopt/config.go
yolocs: This is in the lumberjack _client_ (SDK) running in user's application code. We should add proper prefix to differentiate from other potential app configurations.

myurtoglu (approved):

PR #313, opened 2022-09-27
Reduce the number of tests

sethvargo:
Now that we have a single Go module, this gets a bit simpler :)


sethvargo:
@yolocs I prefer this approach because it ensures all _new_ Go packages are tested by default. The old approach requires us to update CI to test the new packages. 

If we like this approach, I can update the branch protection requirements to prefer these new test names.

bradegler (approved):

yolocs (approved):

PR #314, opened 2022-09-29
Update to latest protos and add scripts

sethvargo:
This updates the protos to the latest HEAD version, re-generates the protos, and adds Makefile targets for both of these tasks, with the ability to expand to other third_party things in the future if needed.


sethvargo:
ETOOBIG. Gonna split up into 2 PRs

Issue #277, opened 2022-08-15
Java examples in docs


yolocs:

yolocs:
https://github.com/abcxyz/lumberjack/blob/main/docs/clients.md#java

PR #319, opened 2022-10-07
docs: add java client set up guide

sqin2019:
#277 
1. java client does not support custom processors, discussed with Chen that we will support that when requested.
2.  used md format to format the files touched.

### DIFF
--- a/docs/clients.md
+++ b/docs/clients.md
@@ -4,13 +4,13 @@
 
 Lumberjack client consists of the following processors and they run in order:
 
-1. Validator(s): To validate the audit log requests
-2. Mutator(s): To mutate the audit log requests, e.g. applying common labels
-3. Backend(s): Specifying where to send the logs.
+1.  Validator(s): To validate the audit log requests
+2.  Mutator(s): To mutate the audit log requests, e.g. applying common labels
+3.  Backend(s): Specifying where to send the logs.
 
 You can use a [config file](./config.md) to ease the client initialization, or
-"assemble" a client by yourself in code if you are onboarding to Go client.
-See exampels below.
+"assemble" a client by yourself in code if you are onboarding to Go client. See
+exampels below.
 
 ## Go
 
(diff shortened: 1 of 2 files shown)


docs/clients.md
```diff
@@ -108,4 +109,23 @@ if err != nil {
...
+
+// For gRPC services, use AuditLoggingServerInterceptor.
+AuditLoggingServerInterceptor interceptor =
+        injector.getInstance(AuditLoggingServerInterceptor.class);
+
+// Otherwise, use LoggingClient directly.
+LoggingClient client = injector.getInstance(LoggingClient.class);
+```
+
+### Extend
```
yolocs, line 129: Let's have the same subsections as Go. But under "### "Assemble" a client in code" we say what you say at the top: "It is not supported..."

docs/clients.md
```diff
@@ -108,4 +109,23 @@ if err != nil {
 
 ## Java
 
-TODO
+It is not recommended and supported to "assemble" a client in code,
```
yolocs, line 112:
```suggestion
Currently we don't support "assembling" a client in Java code. Please file an issue if you require such capability.
```

yolocs: And move to the other subsection

docs/clients.md
```diff
@@ -108,4 +109,23 @@ if err != nil {
...
+please use config file. The default config path is `audit_logging.yml`, or
+if it is overwritten via env var, AUDIT_CLIENT_CONFIG_PATH.
```
yolocs, line 114:
```suggestion
The default config path is `audit_logging.yml`. You can override it with env var `AUDIT_CLIENT_CONFIG_PATH`.
```

yolocs: Move this to "### Create a client from a config file"

docs/clients.md
```diff
@@ -108,4 +109,23 @@ if err != nil {
...
+
+### Create a client from a config file
+
+```java
```
yolocs, line 118: Should we provide a intro to Guice, e.g. a link?

docs/clients.md
```diff
@@ -108,4 +109,23 @@ if err != nil {
...
+AuditLoggingServerInterceptor interceptor =
+        injector.getInstance(AuditLoggingServerInterceptor.class);
+
+// Otherwise, use LoggingClient directly.
+LoggingClient client = injector.getInstance(LoggingClient.class);
+```
+
+### Extend
+
+Extending with custom processors is not supported at the moment, please file an issue if needed.
```
yolocs, line 131:
```suggestion
Due to all the log processors are initialized via injection, we currently don't support vending custom log processors to the audit log client. Please file an issue if you need such capability.
```

yolocs (changes requested):
Some nits

sethvargo (approved):

yolocs (approved):

PR #325, opened 2022-10-18
chore: update cleanup workflow to clean up images in the new location

yolocs:
Follow up of https://github.com/abcxyz/lumberjack/pull/322
Now all the CI images are stored in the same registry with the commit being the tag (40 characters long).


bradegler (approved):
👍

Issue #286, opened 2022-09-01
Consider adopting goreleaser


sethvargo:
### TL;DR

As we publish binaries and tag versions, using [Goreleaser](https://goreleaser.com/) will make our lives a bit easier, especially since we need to compile multiple binaries.

### Detailed design

Make a config file like https://github.com/GoogleCloudPlatform/berglas/blob/main/.goreleaser.yml

Then set up a GitHub Action like https://github.com/GoogleCloudPlatform/berglas/blob/main/.github/workflows/release.yml

We would use sigstore instead of GPG signing, but the rest of the configuration is still probably valid.

PR #328, opened 2022-10-21
chore: add goreleaser and use it in CI

yolocs:
Fix: https://github.com/abcxyz/lumberjack/issues/286

We use goreleaser in CI to publish the ingestion service image so we can keep it close to how we will actually release.
It doesn't make sense to use goreleaser for CI test clients image build since we don't really release those images.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -14,16 +14,36 @@ See sample log sinks in Terraform for application-level audit logs in
 `google_logging_project_sink` resources
 [here](../terraform/modules/server-sink/main.tf).
 
-In case you want all your _cloud_ audit logs to end up in the same log storage,
+In case you want all your *cloud* audit logs to end up in the same log storage,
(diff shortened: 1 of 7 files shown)


docs/setup.md
```diff
@@ -14,16 +14,36 @@ See sample log sinks in Terraform for application-level audit logs in
...
 ## (Optional) Ingestion Service
 
-Build the server with [Dockerfile](../scripts/server.dockerfile). Where you run
-the ingestion server doesn't matter as long as it's accessible by the
-applications. See [sample deployment](../terraform/modules/server-service/) with
-Cloud Run in Terraform.
+To build the ingestion service:
+
+```sh
+# By default we use JVS CI container registry for the images.
```
bradegler, line 26:
```suggestion
# By default we use Lumberjack CI container registry for the images.
```

bradegler (approved):

sethvargo (approved):

PR #340, opened 2022-11-30
feat: Add integration test setup to test ingestion-service-less setup

### LINKED ISSUE #196: Add integration test setup to test ingestion-service-less setup
The existing integration test runs with the ingestion service. We should also test the setup without ingestion service to ensure we can still collect logs correctly.

capri-xiyue:
fixed #196

### DIFF
--- a/terraform/modules/ci/main.tf
+++ b/terraform/modules/ci/main.tf
@@ -19,6 +19,10 @@ locals {
     "AUDIT_CLIENT_BACKEND_REMOTE_ADDRESS" : "${trimprefix(module.server_service.audit_log_server_url, "https://")}:443",
     "AUDIT_CLIENT_CONDITION_REGEX_PRINCIPAL_INCLUDE" : ".*",
   }
+  cl_client_env_vars = {
+    "AUDIT_CLIENT_BACKEND_CLOUDLOGGING_DEFAULT_PROJECT" : "true",
+    "AUDIT_CLIENT_CONDITION_REGEX_PRINCIPAL_INCLUDE" : ".*",
+  }
   short_sha = substr(var.commit_sha, 0, 7)
 }
 
@@ -55,6 +59,37 @@ resource "google_cloud_run_service" "client_services" {
 
         dynamic "env" {
           for_each = local.client_env_vars
+
+          content {
+            name  = env.key
+            value = env.value
+          }
+        }
+      }
+    }
+  }
+}
+
+resource "google_cloud_run_service" "cl_client_services" {
+  for_each = var.client_images
+
+  name     = "${each.key}-${local.short_sha}"
+  project  = var.client_project_id
+  location = var.region
+
+  template {
+    spec {
+      containers {
+        image = each.value
+
+        resources {
+          limits = {
+            memory = "1G"
+          }
+        }
+
+        dynamic "env" {
+          for_each = local.cl_client_env_vars
 
           content {
             name  = env.key
--- a/terraform/modules/ci/outputs.tf
+++ b/terraform/modules/ci/outputs.tf
@@ -21,6 +21,7 @@ output "server_url" {
 output "client_endpoints" {
   value = {
     for key, value in var.client_images :
-    key => lookup(google_cloud_run_service.client_services, key).status[0].url
+    key => concat(lookup(google_cloud_run_service.client_services, key).status[0].url,
+           lookup(google_cloud_run_service.cl_client_services, key).status[0].url)
   }
 }


capri-xiyue:
Somehow the test-ingestion-service-less for Java doesn't work, but for Golang it wroks. I will change the scope of this PR to Golang first and the investigate why Java doesn't work.

capri-xiyue:
@yolocs @sqin2019 This one is ready for review now. Will need to create another PR for Java part. Looks like there is some bug in Java client.

terraform/modules/ci/outputs.tf
```diff
@@ -18,9 +18,16 @@ output "server_url" {
...
+    key => lookup(google_cloud_run_service.ingestion_client_services, key).status[0].url
+  }
+}
+
+output "cl_client_endpoints" {
+  value = {
+    for key, value in var.client_images :
+    key => lookup(google_cloud_run_service.cl_client_services, key).status[0].url
   }
 }
```
yolocs, line 33: Will it be easier in the workflow if we use https://developer.hashicorp.com/terraform/language/functions/merge or https://developer.hashicorp.com/terraform/language/functions/concat to combine the output in `clients_endpoints`?

terraform/modules/ci/main.tf
```diff
@@ -35,10 +39,41 @@ module "server_service" {
   disable_dedicated_sa = true
 }
 
-resource "google_cloud_run_service" "client_services" {
+resource "google_cloud_run_service" "ingestion_client_services" {
```
yolocs, line 42: Since we're making this change, can we make the names more intuitive? Like

`ingestion_backed_services` and `cloudlogging_backed_services`. Especially for `cl` my first idea is changelist rather cloud logging.

capri-xiyue:
Did further investigation.
The java failure is because java client write the audit logs with wrong format.
As https://github.com/abcxyz/lumberjack/blob/fc452bc6893737e8820b3620572e962b32673d63/integration/testrunner/httprunner/http.go#L77 said, for http, we query the message via `labels.trace_id`, but java client(cloud logging only) wrote the filed to `jsonPayload.labels`, see example here 
Screen Shot 2022-12-01 at 10 18 19 AM

For java grpc, as https://github.com/abcxyz/lumberjack/blob/main/integration/testrunner/grpcrunner/grpc.go#L333 said, for grpc, we query the message via `jsonPayload.resource_name`, but java client(cloud logging only) wrote the filed to `jsonPayload.payload.resource_name`, see example here 
Screen Shot 2022-12-01 at 10 21 59 AM

terraform/modules/ci/outputs.tf
capri-xiyue: It is possible, but it makes it more difficulty to debug with some specific tests. For example, I just want to debug with java grpc(cloud logging).

terraform/modules/ci/main.tf
capri-xiyue: done

capri-xiyue:
Changed the scope of PR to golang only since we need to fix the newly detected bug https://github.com/abcxyz/lumberjack/issues/341 before we enable the java part of ingestion-service-less e2e test

capri-xiyue:
I fixed the java client http test error
But for java client grpc test, still have some problems
Somehow it uses `jsonpayload.resourcename` instead of `jsonpayload.resource_name`

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
```diff
@@ -63,9 +63,12 @@ public AuditLogRequest process(AuditLogRequest auditLogRequest) throws LogProces
           LogEntry.newBuilder(
                   Payload.JsonPayload.of(
                       mapper.readValue(
-                          JsonFormat.printer().print(auditLogRequest),
+                          JsonFormat.printer()
+                              .preservingProtoFieldNames()
+                              .print(auditLogRequest.getPayload()),
                           new TypeReference<Map<String, ?>>() {})))
               .setLogName(getLogNameFromLogType(auditLogRequest.getType()))
+              .setLabels(auditLogRequest.getLabelsMap())
```
sqin2019, line 71: Do we also need the operation and timestamp fields to make it identical to go cloud logging processor? https://github.com/abcxyz/lumberjack/blob/545bd8eeb449168cb2b863a3177ae1d1c5cdc3ca/clients/go/pkg/cloudlogging/processor.go#L131

capri-xiyue: I create another bug to track it https://github.com/abcxyz/lumberjack/issues/341. Looks like it is not trivial, need to do some type conversion.

capri-xiyue:
> I fixed the java client http test error But for java client grpc test, still have some problems Somehow it uses `jsonpayload.resourcename` instead of `jsonpayload.resource_name`

Thanks @sqin2019 for helping figuring out the `jsonpayload.resource_name` fix with `preservingProtoFieldNames()`

sqin2019 (approved):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
yolocs: Thank you both for catching the issue!

Can we add a comment here for why we need this?

yolocs (approved):

PR #343, opened 2022-12-06
fix: java client cloud processor log entry inconsistency

sqin2019:
To be consistent with [go client](https://github.com/abcxyz/lumberjack/blob/545bd8eeb449168cb2b863a3177ae1d1c5cdc3ca/clients/go/pkg/cloudlogging/processor.go#L128)
```
	logEntry := logging.Entry{
		Payload:   logReq.Payload,
		Labels:    logReq.Labels,
		Operation: logReq.Operation,
		Timestamp: logReq.Timestamp.AsTime(),
	}
```
Where added fields are set:
[Operation](https://github.com/abcxyz/lumberjack/blob/563a9dd75e25d3645e6bebecc506d7770679f17c/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java#L125),
[timestamp](https://github.com/abcxyz/lumberjack/blob/563a9dd75e25d3645e6bebecc506d7770679f17c/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java#L233)


clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
```diff
@@ -69,35 +64,22 @@ public AuditLogRequest process(AuditLogRequest auditLogRequest) throws LogProces
...
               .setLabels(auditLogRequest.getLabelsMap())
-              .setResource(MonitoredResource.newBuilder(MONITORED_RESOURCE_TYPE).build())
+              .setOperation(
+                  Operation.of(
+                      auditLogRequest.getOperation().getId(),
+                      auditLogRequest.getOperation().getProducer()))
+              .setTimestamp(
+                  Instant.ofEpochSecond(
+                      auditLogRequest.getTimestamp().getSeconds(),
+                      auditLogRequest.getTimestamp().getNanos()))
```
myurtoglu, line 75: Are these types incompatible? Couldn't we just do:

```
.setOperation(auditLogRequest.getOperation())
.setTimestamp(auditLogRequest.getTimestamp())
```

myurtoglu (approved):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
sqin2019: They are incompatible:
setOperation needs Operation whereas auditLogRequest.getOperation() is LogEntryOperation 
setTimestamp needs Instant whereas auditLogRequest.getTimestamp() is google.protobuf.Timestamp

myurtoglu: Ack, makes sense.

For the timestamp, [using this](https://github.com/googleapis/java-logging/blob/059d62acd81190489122490cc072d0d0cb5e8cd8/google-cloud-logging/src/main/java/com/google/cloud/logging/JavaTimeConversions.java#L51) may be helpful.

sqin2019: Seems it is not visible outside of that package, so keeping the way it is.

sqin2019:
revert back, since it does not help. Currently from bigquery logs, below are the diffs:
```
# java client cloud logging
  "resource": {
    "type": "global",
    "labels": {
      "location": null,
      "service_name": null,
      "project_id": "github-ci-app-0",
      "configuration_name": null,
      "revision_name": null
    }
  },
  
# go client cloud logging
"resource": {
    "type": "cloud_run_revision",
    "labels": {
      "location": "us-central1",
      "service_name": "ljc-go-http-c7c9163-cloudlogging",
      "project_id": "github-ci-app-0",
      "configuration_name": "ljc-go-http-c7c9163-cloudlogging",
      "revision_name": "ljc-go-http-c7c9163-cloudlogging-47zql"
    }
  },
```

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
```diff
@@ -71,6 +73,14 @@ public AuditLogRequest process(AuditLogRequest auditLogRequest) throws LogProces
                           new TypeReference<Map<String, ?>>() {})))
               .setLogName(getLogNameFromLogType(auditLogRequest.getType()))
               .setLabels(auditLogRequest.getLabelsMap())
+              .setOperation(
+                  Operation.of(
+                      auditLogRequest.getOperation().getId(),
```
capri-xiyue, line 78: I saw that `operation` object also has `first` and `last` field

sqin2019: Yes, we only need id and producer according to here: https://github.com/abcxyz/lumberjack/blob/563a9dd75e25d3645e6bebecc506d7770679f17c/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java#L125

capri-xiyue: Can we add a code comment to explain why we only need these two fields?

yolocs, line 78: This is weird. Both https://github.com/abcxyz/lumberjack/blob/563a9dd75e25d3645e6bebecc506d7770679f17c/protos/v1alpha1/audit_log_request.proto#L82 and https://github.com/googleapis/googleapis/blob/master/google/logging/v2/log_entry.proto#L154 use `LogEntryOperation`. But in Java, `setOperation` expects an `Operation` instead?

sqin2019: Yes, will do.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
```diff
@@ -71,6 +73,14 @@ public AuditLogRequest process(AuditLogRequest auditLogRequest) throws LogProces
...
+                      auditLogRequest.getOperation().getProducer()))
+              .setTimestamp(
```
yolocs, line 80: What if timestamp and operation are null in the `auditLogRequest`? Will we get an exception?

sqin2019: Solid point, but it didn't for the java http client case in integration, the bigquery log just shows null for operation. checking...

sqin2019: Null cases are covered by protobuf, it is in the genrated code, let me see if I can find a link.

yolocs (approved):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
sqin2019: Yes, the LogEntry here is different then the log_entry proto, it has a internal/package level method to convert log entry proto.

myurtoglu: Not the cleanest thing maybe, but both for Operation and for Timestamp, after our review round yesterday, given those methods seem slightly more generalized than what we are doing here, I thought maybe we should verbatim copy/paste those two internal methods and use them as helpers instead of manually doing the conversion inline?

sqin2019: Changed to add all fields to cover http client case.

capri-xiyue (approved):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
sqin2019: Thanks for calling this out! I quickly checked the repo and didn't see other places that need the conversion helpers, and probably unlikely in future as well since we only need this LogEntry to write to cloud logging for java. We can do the refactor when needed, what do you think?

myurtoglu: Sure, either way fine by me.

PR #345, opened 2022-12-08
feat: add more fields validation in e2e test

### LINKED ISSUE #344: Lumberjack integ test validates more fields
### TL;DR

Lumberjack integ test validates more fields

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

capri-xiyue:
fixed #344

### DIFF
--- a/integration/testrunner/httprunner/http.go
+++ b/integration/testrunner/httprunner/http.go
@@ -74,6 +74,9 @@ func TestHTTPEndpoint(ctx context.Context, tb testing.TB, endpointURL string,
 
 func makeQueryForHTTP(client bigquery.Client, id string, projectID string, datasetQuery string) *bigquery.Query {
 	// Cast to int64 because the result checker expects a number.
-	queryString := fmt.Sprintf("SELECT CAST(EXISTS (SELECT * FROM `%s.%s` WHERE labels.trace_id=?) AS INT64)", projectID, datasetQuery)
+	queryString := fmt.Sprintf("SELECT CAST(EXISTS (SELECT * FROM `%s.%s` WHERE labels.trace_id=?", projectID, datasetQuery)
+	queryString += ` AND jsonPayload.service_name IS NOT NULL`
+	queryString += ` AND jsonPayload.authentication_info.principal_email IS NOT NULL`
+	queryString += ") AS INT64)"
 	return utils.MakeQuery(client, id, queryString)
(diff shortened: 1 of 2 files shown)


sqin2019 (approved):

yolocs (approved):
Thanks!

integration/testrunner/httprunner/http.go
```diff
@@ -74,6 +74,9 @@ func TestHTTPEndpoint(ctx context.Context, tb testing.TB, endpointURL string,
 
 func makeQueryForHTTP(client bigquery.Client, id string, projectID string, datasetQuery string) *bigquery.Query {
 	// Cast to int64 because the result checker expects a number.
-	queryString := fmt.Sprintf("SELECT CAST(EXISTS (SELECT * FROM `%s.%s` WHERE labels.trace_id=?) AS INT64)", projectID, datasetQuery)
+	queryString := fmt.Sprintf("SELECT CAST(EXISTS (SELECT * FROM `%s.%s` WHERE labels.trace_id=?", projectID, datasetQuery)
+	queryString += ` AND jsonPayload.service_name IS NOT NULL`
+	queryString += ` AND jsonPayload.authentication_info.principal_email IS NOT NULL`
+	queryString += ") AS INT64)"
```
sethvargo, line 80: Long term, I'm not sure this is the best way to test this. We should probably search for the trace, serialize the result into a structure, they assert the structure as a diff.

Right now, if the `principle_email` is missing, this will return NOT_FOUND, but it's not immediately clear to the test runner _why_ it's not found (there's 4 search criteria).

capri-xiyue: Created a GitHub issue to track a better fix(https://github.com/abcxyz/lumberjack/issues/346).
I agree that currently we have to take a look at the data in BQ to see why it's not found and it is not quite straight forward. But  serialize the result into a structure and then assert the whole structure as a diff could be too much. A lot of fields related to `justification` in the BQ data are quite complex. I think we can serialize the result into a structure and assert some essential fields like `jsonPayload.service_name` and `jsonPayload.authentication_info.principal_email`

sethvargo: Right - I don't think we'd assert the _entire_ structure, but we can assert specific fields and return more specific errors when things don't match.

PR #348, opened 2023-01-04
chore: support using Go generate to generate proto code

yolocs:


Makefile
```diff
@@ -14,11 +14,11 @@
 
 # protoc compiles the protobufs.
 protoc:
```
sethvargo, line 16: Should we drop the Makefile target then?

sethvargo (approved):

Makefile
yolocs: Can do! (But now the other makefile target is so lonely :)

PR #349, opened 2023-01-04
Fix up some linters

sethvargo:


clients/go/pkg/auditopt/config.go
```diff
@@ -99,7 +99,9 @@ func InterceptorFromConfigFile(ctx context.Context, path string) audit.Intercept
 func interceptorFromConfigFile(ctx context.Context, path string, lookuper envconfig.Lookuper) audit.InterceptorOption {
 	return func(i *audit.Interceptor) error {
 		fc, err := os.ReadFile(path)
-		if err != nil {
+		// We ignore ErrNotExist when reading the file because we
```
sethvargo, line 102: This was likely a bug, since we ignore EOF in the other config loading.

integration/testrunner/db.go
```diff
@@ -57,14 +57,29 @@ func queryIfAuditLogExists(ctx context.Context, tb testing.TB, query *bigquery.Q
...
 	bqQuery := bqClient.Query(queryString)
 	bqQuery.Parameters = []bigquery.QueryParameter{{Value: id}}
 	return bqQuery
 }
 
-func makeBigQueryClient(ctx context.Context, projectID string) (*bigquery.Client, error) {
-	return bigquery.NewClient(ctx, projectID)
+// makeBigQueryClient creates a new client and automatically closes the
+// connection when the tests finish.
+func makeBigQueryClient(ctx context.Context, tb testing.TB, projectID string) *bigquery.Client {
```
sethvargo, line 68: Made this a bit nicer by accepting a tb, so we can handle closing for callers.

integration/testrunner/main_test.go
```diff
@@ -68,6 +68,7 @@ func TestHTTPEndpoints(t *testing.T) {
 	}
 
 	for i, tc := range tests {
+		i := i
```
sethvargo, line 71: Twas a bug

clients/go/test/grpc-app/main.go
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-// https://www.apache.org/licenses/LICENSE-2.0
+//     http://www.apache.org/licenses/LICENSE-2.0
```
myurtoglu, line 7: Why (curiosity)? The website itself self-links with https ?

myurtoglu (approved):

clients/go/test/grpc-app/main.go
sethvargo: IANAL, but using `https` could technically invalidate the license: https://www.apache.org/licenses/LICENSE-2.0#apply

I'm not sure it would hold up in court, but that link above is the license body we need to apply.

PR #353, opened 2023-01-05
Update Java deps

sethvargo:
I went through the changelogs and didn't see anything crazy from springboot 2->3 and the tests are passing, but I'm far from a Java expert.

### DIFF
--- a/clients/java-logger/grpc-test-app/pom.xml
+++ b/clients/java-logger/grpc-test-app/pom.xml
@@ -145,6 +145,11 @@
       <version>1.4.5</version>
     </dependency>
     <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-core</artifactId>
+      <version>1.4.5</version>
+    </dependency>
+    <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-auth</artifactId>
       <version>1.51.1</version>
--- a/clients/java-logger/grpc-test-app/pom.xml
+++ b/clients/java-logger/grpc-test-app/pom.xml
@@ -13,6 +13,17 @@
   <groupId>com.abcxyz.lumberjack</groupId>
   <version>0.0.1</version>
 
+  <properties>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
+    <java.version>11</java.version>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+    <lumberjack.basedir>../../..</lumberjack.basedir>
+    <protobuf.version>3.21.7</protobuf.version>
+    <protoc.version>3.21.12</protoc.version>
+  </properties>
+
   <dependencyManagement>
     <dependencies>
       <dependency>
@@ -24,16 +35,6 @@
       </dependency>
     </dependencies>
   </dependencyManagement>
-
-  <properties>
-    <maven.compiler.source>11</maven.compiler.source>
-    <maven.compiler.target>11</maven.compiler.target>
-    <java.version>11</java.version>
-    <lumberjack.basedir>../../..</lumberjack.basedir>
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <protobuf.version>3.19.1</protobuf.version>
-    <protoc.version>3.19.1</protoc.version>
-  </properties>
 
   <dependencies>
     <dependency>
@@ -55,7 +56,7 @@
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
-      <version>3.21.12</version>
+      <version>${protobuf.version}</version>
     </dependency>
     <dependency>
       <groupId>com.google.api.grpc</groupId>
@@ -207,6 +208,7 @@
               <goal>run</goal>
             </goals>
             <configuration>
+              <protocVersion>${protoc.version}</protocVersion>
               <includeStdTypes>true</includeStdTypes>
               <inputDirectories>
                 <include>${lumberjack.basedir}/internal/protos/talker</include>
@@ -218,8 +220,7 @@
                 </outputTarget>
                 <outputTarget>
                   <type>grpc-java</type>
-                  <pluginArtifact>io.grpc:protoc-gen-grpc-java:1.41.0
-                  </pluginArtifact>
+                  <pluginArtifact>io.grpc:protoc-gen-grpc-java:1.50.1</pluginArtifact>
                 </outputTarget>
               </outputTargets>
             </configuration>
@@ -230,6 +231,7 @@
               <goal>run</goal>
             </goals>
             <configuration>
+              <protocVersion>${protoc.version}</protocVersion>
               <includeStdTypes>true</includeStdTypes>
               <includeDirectories>
                 <include>${lumberjack.basedir}/internal/protos/talker</include>
--- a/clients/java-logger/grpc-test-app/pom.xml
+++ b/clients/java-logger/grpc-test-app/pom.xml
@@ -18,7 +18,7 @@
       <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>libraries-bom</artifactId>
-        <version>26.1.1</version>
+        <version>26.2.0</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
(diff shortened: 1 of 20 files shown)


myurtoglu:
> I went through the changelogs and didn't see anything crazy from springboot 2->3 and the tests are passing, but I'm far from a Java expert.

I'm far from being a Java (and especially Spring) expert as well, that said, [checking here](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#review-system-requirements), it says Java version 17 is needed, whereas in [here](https://github.com/abcxyz/lumberjack/blob/ce0b8920bd04842a3b373f180a50f27d69a0ce7c/clients/java-logger/shell/pom.xml#L16), we have version 11 set. Speaking of the same file, I wonder if [this](https://github.com/abcxyz/lumberjack/blob/ce0b8920bd04842a3b373f180a50f27d69a0ce7c/clients/java-logger/shell/pom.xml#L9) should also be bumped to something 3.X. On the bright side, I don't see anything else problematic on the migration guide for us. So, maybe try bumping those two as well and let's see if all tests pass?

myurtoglu (approved):
Preemptively approving, presuming bumping Java to 17 and `spring-boot-starter-parent` to 3.0.1 will do the trick 🤞

clients/java-logger/grpc-test-app/pom.xml
```diff
@@ -136,18 +136,18 @@
...
-      <version>2.7.3</version>
+      <version>3.0.1</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
-      <version>1.2.3</version>
+      <version>1.4.5</version>
     </dependency>
```
yolocs, line 146: Looking at https://stackoverflow.com/questions/9717518/java-lang-nosuchmethoderror-ch-qos-logback-classic-loggercontext-getbithtime, I wonder if we need to explicit add `logback-core` with the same `logback-classic` version. @myurtoglu

myurtoglu: Yeah, along the same lines, doing some Googling, I saw that the missing method [was introduced in version 1.30](https://logback.qos.ch/apidocs/ch/qos/logback/core/util/EnvUtil.html#logbackVersion()) but [we also have another dependency that still uses 1.2.11](https://github.com/googleapis/java-logging-logback/blob/883b22c051f83598d3732d6a59972c6b4e9765fb/pom.xml#L17).

myurtoglu: > Looking at https://stackoverflow.com/questions/9717518/java-lang-nosuchmethoderror-ch-qos-logback-classic-loggercontext-getbithtime, I wonder if we need to explicit add `logback-core` with the same `logback-classic` version. @myurtoglu

And this I don't know, [but classic seems to include core anyway](https://github.com/qos-ch/logback/blob/bba7a908c87a4a483860f6d49695ec2348c56e49/logback-classic/pom.xml#L26).

sethvargo:
Okay - I removed the Spring upgrade but kept everything else. I also found a bunch of type warning assertions and a pretty bad bug: https://github.com/abcxyz/lumberjack/issues/354

clients/java-logger/grpc-test-app/pom.xml
myurtoglu: Resolving conversation to unblock this change.

yolocs:
I failed to _locate_ the root cause but I'm pretty sure it's some change introduced in grpc-java that changed how exceptions are propagated to the interceptor. In some cases (e.g. `onError(exception)`), it triggers `call.close(...)` directly with the error code. Sometimes (throw exception in unary method), the exception propagates to `onHalfClose()`. But this is conclusion from _many many_ trail and errors. My java knowledge doesn't allow me to go any deeper without pulling my eyes out.

So my resolution is to convert all explicit exceptions and redirect them to `call.close(...)`. And use `call.close(...)` to be the ultimate place to capture all error code and logs the final audit log before close the call. It proves to be effective. Here is a sample final log in a stream (that was failing). The last request's value is 4 which triggered the server to throw an exception. And its corresponding code is captured in the status (`INVALID_ARGUMENT`).

```
{
  "insertId": "1ra0t1vg1mnqq52",
  "jsonPayload": {
    "request": {
      "value": 4,
      "target": "f10db6ce-5100-4888-b17d-0fc316197ce8"
    },
    "service_name": "abcxyz.test.Talker",
    "resource_name": "f10db6ce-5100-4888-b17d-0fc316197ce8",
    "authentication_info": {
      "principal_email": "gh-access-sa@lumberjack-dev-infra.iam.gserviceaccount.com"
    },
    "method_name": "abcxyz.test.Talker/FailOnFour",
    "status": {
      "code": 3,
      "message": "INVALID_ARGUMENT"
    }
  }
}
```

Along the way, I split the gRPC tests into subtests and made them parallel. That'd probably shorten the integration testing for like 10-20s :)
Also found some issues we should fix: https://github.com/abcxyz/lumberjack/issues/360 and https://github.com/abcxyz/lumberjack/issues/361

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/AuditLoggingServerInterceptor.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Lumberjack authors (see AUTHORS file)
+ * Copyright 2023 Lumberjack authors (see AUTHORS file)
```
sethvargo, line 2: Nit: We shouldn't actually update copyrights. Only on _new_ files.

integration/testrunner/grpcrunner/grpc.go
```diff
@@ -73,32 +73,34 @@ var (
 // TestGRPCEndpoint runs tests against a GRPC endpoint. The given GRPC must
 // define a projectID and datasetQuery. If a TalkerClient or BigQueryClient are
 // not provided, they are instantiated via the defaults.
-func TestGRPCEndpoint(ctx context.Context, tb testing.TB, g *GRPC) {
-	tb.Helper()
+//
+//nolint:tparallel // t.Parallel() is already called
+func TestGRPCEndpoint(ctx context.Context, t *testing.T, g *GRPC) {
```
sethvargo, line 78: If we name this lowercase testGRPCEndpoint, then you don't need the nolint. Test helpers should also accept testing.TB, not *testing.T.

sethvargo: (Throughout)

integration/testrunner/grpcrunner/grpc.go
```diff
@@ -320,7 +344,7 @@ func createConnection(tb testing.TB, addr string, idToken string) *grpc.ClientCo
 	addr = addr + ":443"
 	conn, err := grpc.Dial(addr, grpc.WithPerRPCCredentials(rpcCreds), grpc.WithTransportCredentials(creds))
```
sethvargo, line 345: Can be a followup, but we should be using DialContext instead: https://pkg.go.dev/google.golang.org/grpc#DialContext

integration/testrunner/grpcrunner/grpc.go
```diff
@@ -111,21 +113,40 @@ func TestGRPCEndpoint(ctx context.Context, tb testing.TB, g *GRPC) {
...
 	}
 
-	g.runHelloCheck(ctx, tb)
-	g.runFailCheck(ctx, tb)
-	g.runFibonacciCheck(ctx, tb)
-	g.runAdditionCheck(ctx, tb)
-	g.runFailOnFourCheck(ctx, tb)
+	t.Run("hello_req_unary_success", func(t *testing.T) {
+		t.Parallel()
+		g.runHelloCheck(ctx, t)
```
sethvargo, line 128: Should we just inline these?

integration/testrunner/grpcrunner/grpc.go
sethvargo: > This is in a different package so the func must be exposed. How about `CheckGRPCEndpoint`?

I see, so we call this from other packages? Can it be inlined instead?

integration/testrunner/grpcrunner/grpc.go
yolocs: Done

integration/testrunner/grpcrunner/grpc.go
yolocs: Done. I also removed the top `t.Helper()` in this file. Here locates the real testing logic rather than a test helper. I found it's hard to locate the line of error if I'm putting `t.Helper()` because it gave me the line in the `main_test.go`

PR #357, opened 2023-01-06
Upgrade protoc and proto-gen-java

sethvargo:
This fixes an issue where it's not possible to run on an M1 Mac because amd binaries are not available on the older versions.

### DIFF
--- a/clients/java-logger/grpc-test-app/pom.xml
+++ b/clients/java-logger/grpc-test-app/pom.xml
@@ -13,6 +13,17 @@
   <groupId>com.abcxyz.lumberjack</groupId>
   <version>0.0.1</version>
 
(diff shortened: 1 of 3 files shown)


capri-xiyue (approved):

sethvargo:
@myurtoglu @yolocs since you're also working on upgrades. I'm _hoping_ this makes it so I can actually run locally again 😄

myurtoglu (approved):

myurtoglu:
> @myurtoglu @yolocs since you're also working on upgrades. I'm _hoping_ this makes it so I can actually run locally again 😄

Sure 😄

yolocs (approved):

PR #362, opened 2023-01-10
Test out new CI flows

sethvargo:


.github/workflows/ci.yml
```diff
@@ -32,10 +32,12 @@ jobs:
       directory: 'terraform'
       terraform_version: '1.2'
 
+  # TODO(sethvargo): fix branch and remove golangci_url
   go_lint:
-    uses: 'abcxyz/pkg/.github/workflows/go-lint.yml@main' # ratchet:exclude
+    uses: 'abcxyz/pkg/.github/workflows/go-lint.yml@sethvargo/lint' # ratchet:exclude
     with:
       go_version: '1.19'
+      golangci_url: 'https://raw.githubusercontent.com/abcxyz/pkg/fddbebf/.golangci.yml'
```
sethvargo, line 40: This is still an open AI because it depends on https://github.com/abcxyz/pkg/pull/50

clients/go/apis/v1alpha1/config.go
```diff
@@ -17,7 +17,7 @@ package v1alpha1
 import (
 	"fmt"
 
-	"go.uber.org/multierr"
+	"github.com/hashicorp/go-multierror"
```
sethvargo, line 20: I needed to switch because we need a way to check if the multierror is "nil" before wrapping.

.github/workflows/ci.yml
verbanicm: LGTM, other PR is approve if you wanna merge that and then fix this.

verbanicm (approved):

clients/go/apis/v1alpha1/config.go
yolocs, line 20: Somewhere I saw a comment about check nil err. I think the old package can do that as well. But it's good we standardize on one lib for multi error handling.

yolocs (approved):
Should we just merge https://github.com/abcxyz/pkg/pull/50 and update this PR?

sethvargo:
> Should we just merge https://github.com/abcxyz/pkg/pull/50 and update this PR?

Yea - I want to do this + the JVS one + pkg at the same time.

clients/go/apis/v1alpha1/config.go
sethvargo: Can you add it to our running doc?

yolocs: Already did!

Issue #203, opened 2022-03-24
Fix inconsistency in config loading between Java and Go


yolocs:
For a sample config

```yaml
backend:
  remote:
    address: "" # empty
```

In Go, it will return an error saying `address` is required for a remote backend config.
In Java, it won't return an error; instead, it will treat that as a remote backend is not needed (enabled).

capri-xiyue:
Do we need to fix such inconsistency?

sethvargo:
I agree with @capri-xiyue - can we fix these inconsistencies instead of just documenting them?

yolocs:
Yes, it would be preferable to completely fix the inconsistency. But there are 1) technical viability 2) language idiomaticity to consider. For this one, I agree we should just fix it. Renaming the title.

sqin2019:
In addition to remote backend config, other inconsistencies: 
- cloudlogging set default logic
- locallogging missing in go side
- go side missing LogType Unspecified
- when security_context is not nil, go client require nonnil and nonempty security_context key where java only require nonnil security_context.
- not sure if [regex can be nil](https://github.com/abcxyz/lumberjack/blob/1cb35ef2faa85dda5a5d7f487e1272c2c66fd539/clients/go/apis/v1alpha1/config.go#L243) even when filter related env vars are set.

PR #373, opened 2023-02-09
fix: add allowBreakglass in both go and java client

sqin2019:
part of #203 
tested the env var locally.

### DIFF
--- a/clients/go/apis/v1alpha1/config.go
+++ b/clients/go/apis/v1alpha1/config.go
@@ -367,6 +367,10 @@ type Justification struct {
 	// all manual calls are expected to pass in a justification in the "justification-token" header.
 	// If omitted, justifications will not be added to logs, even if provided.
 	Enabled bool `yaml:"enabled,omitempty" env:"JUSTIFICATION_ENABLED,overwrite"`
+
+	// AllowBreakglass indicates whether JVS should allow unsigned justification.
+	// If justification is not enabled, AllowBreakglass will be ignored, even if provided.
+	AllowBreakglass bool `yaml:"allow_breakglass,omitempty" env:"JUSTIFICATION_ALLOW_BREAKGLASS,overwrite"`
 }
 
 // Validate validates the Justification.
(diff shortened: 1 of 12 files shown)


clients/go/apis/v1alpha1/config.go
```diff
@@ -367,6 +367,10 @@ type Justification struct {
 	// all manual calls are expected to pass in a justification in the "justification-token" header.
 	// If omitted, justifications will not be added to logs, even if provided.
 	Enabled bool `yaml:"enabled,omitempty" env:"JUSTIFICATION_ENABLED,overwrite"`
+
+	// AllowBreakglass indicates whether JVS should allow unsigned justification.
+	// If justification is not enabled, AllowBreakglass will be ignored, even if provided.
```
sethvargo, line 372: Please document the default behavior.

clients/go/apis/v1alpha1/config.go
```diff
@@ -367,6 +367,10 @@ type Justification struct {
 	// all manual calls are expected to pass in a justification in the "justification-token" header.
 	// If omitted, justifications will not be added to logs, even if provided.
 	Enabled bool `yaml:"enabled,omitempty" env:"JUSTIFICATION_ENABLED,overwrite"`
+
+	// AllowBreakglass indicates whether JVS should allow unsigned justification. If omitted, it is set to false.
```
yolocs, line 371:
```suggestion
	// AllowBreakglass indicates whether JVS should allow unsigned justification. By default, it's false.
```
nit

yolocs (commented):
Don't we need to modify https://github.com/abcxyz/lumberjack/blob/main/clients/go/pkg/justification/processor.go to allow/disallow breakglass?

sqin2019:
> Don't we need to modify https://github.com/abcxyz/lumberjack/blob/main/clients/go/pkg/justification/processor.go to allow/disallow breakglass?

this logic is in JVS client, https://github.com/abcxyz/jvs/blob/74efb454b360cfc45e6564f5c780883ed3c4209a/client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java#L86

yolocs:
I see!

yolocs (approved):

Issue #372, opened 2023-02-09
Add justification token field in log request 


yolocs:
The justification token is passed via log request context at the moment. See https://github.com/abcxyz/lumberjack/blob/main/protos/v1alpha1/audit_log_request.proto#L86-L89 and https://github.com/abcxyz/lumberjack/blob/main/clients/go/pkg/justification/processor.go. This now seems to be overcomplicated.

We can replace that with just a string field in the log request `justification_token`.

PR #376, opened 2023-02-09
Store justification in a dedicated proto field

sethvargo:
This changes the behavior to store the justification token on a new proto field instead of in the untyped and unstructured `Context`.

Fixes https://github.com/abcxyz/lumberjack/issues/372


yolocs (approved):
Now I need to follow up with the java change

PR #377, opened 2023-02-10
fix: precondition failed error should not cause error even in fail-cl…

yolocs:
…ose mode

Also removed "internal error" in Go because any error other than concrete error types are internal error.

### DIFF
--- a/clients/go/pkg/audit/client_test.go
+++ b/clients/go/pkg/audit/client_test.go
@@ -136,7 +136,7 @@ func TestLog(t *testing.T) {
 				testutil.WithMode(api.AuditLogRequest_BEST_EFFORT)),
 		},
 		{
-			name:   "failed_precondition_in_mutator_should_return_nil_on",
+			name:   "failed_precondition_in_mutator_should_return_nil",
 			logReq: testutil.NewRequest(),
 			opts: []Option{
 				WithMutator(testOrderProcessor{name: "fake", returnErr: fmt.Errorf("fake error: %w", auditerrors.ErrPreconditionFailed)}),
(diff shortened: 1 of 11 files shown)


sethvargo (approved):

sqin2019:
Should we also change https://github.com/abcxyz/lumberjack/blob/ea786ece7278998e5f1a797cc206a39385ea3081/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/ValidationProcessor.java#L32 ?

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/PreconditionFailedException.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2021 Lumberjack authors (see AUTHORS file)
```
sqin2019, line 2: nit: 2023

clients/go/pkg/audit/client_test.go
```diff
@@ -136,15 +136,15 @@ func TestLog(t *testing.T) {
 				testutil.WithMode(api.AuditLogRequest_BEST_EFFORT)),
 		},
 		{
-			name:   "failed_precondition_in_mutator_should_return_nil_on_best_effort",
+			name:   "failed_precondition_in_mutator_should_return_nil_on",
```
sqin2019, line 139: nit: failed_precondition_in_mutator_should_return_nil

sqin2019 (approved):

PR #382, opened 2023-02-15
feat: Figured out a better way to verify the e2e test

### LINKED ISSUE #346: Figure out a better way to verify the e2e test
### TL;DR

Previously, we rely on single filed to verify the e2e test, see details here: https://github.com/abcxyz/lumberjack/pull/340#issuecomment-1334172512

To make the e2e test more reliable, we merged https://github.com/abcxyz/lumberjack/pull/345#pullrequestreview-1210804849 to rely on multiple fields to verify the e2e test.

@sethvargo added a comment in https://github.com/abcxyz/lumberjack/pull/345#discussion_r1043842668, we can try to figure out a better way to verify the e2e test. I think to serialize the result into a structure and then assert the whole structure as a diff could be too much. A lot of fields related to `justification` in the BQ data are quite complex. I think we can serialize the result into a structure and assert some essential fields like `jsonPayload.service_name` and `jsonPayload.authentication_info.principal_email`

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

capri-xiyue:
1. fixed #346 
2. added jsonPayload.methodName in go shell app
3.  cleaned BQ schema in test infra project


capri-xiyue:
This PR is just for test. Closed it because PR with same code https://github.com/abcxyz/lumberjack/pull/368 has been merged

PR #383, opened 2023-02-15
feat: added timestamp field validation in e2e and fixed missing timestamp bug in java src

### LINKED ISSUE #381: Make test log consistent
### TL;DR

See context 
https://github.com/abcxyz/lumberjack/pull/368#discussion_r1103233766

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

capri-xiyue:
part of #381 
1. added timestamp in java logging shell
2. not override `resource` field when generating logentry in java
3. added `timestamp` field validation

### DIFF
--- a/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
+++ b/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
@@ -23,7 +23,6 @@ import com.abcxyz.lumberjack.v1alpha1.AuditLogRequestProto;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.cloud.MonitoredResource;
 import com.google.cloud.logging.LogEntry;
 import com.google.cloud.logging.Logging;
 import com.google.cloud.logging.LoggingException;
@@ -47,7 +46,6 @@ import lombok.extern.java.Log;
 @Log
 @AllArgsConstructor(access = AccessLevel.PRIVATE, onConstructor = @__({@Inject}))
 public class CloudLoggingProcessor implements LogBackend {
-  private static final String MONITORED_RESOURCE_TYPE = "global";
   private final ObjectMapper mapper;
   private final Logging logging;
 
@@ -84,7 +82,6 @@ public class CloudLoggingProcessor implements LogBackend {
                   Instant.ofEpochSecond(
                       auditLogRequest.getTimestamp().getSeconds(),
                       auditLogRequest.getTimestamp().getNanos()))
-              .setResource(MonitoredResource.newBuilder(MONITORED_RESOURCE_TYPE).build())
               .build();
       logging.write(Collections.singleton(entry));
       return auditLogRequest;
--- a/clients/java-logger/shell/src/main/java/com/abcxyz/lumberjack/loggingshell/LoggingController.java
+++ b/clients/java-logger/shell/src/main/java/com/abcxyz/lumberjack/loggingshell/LoggingController.java
@@ -22,6 +22,9 @@ import com.abcxyz.lumberjack.v1alpha1.AuditLogRequest;
 import com.abcxyz.lumberjack.v1alpha1.AuditLogRequest.LogType;
 import com.google.cloud.audit.AuditLog;
 import com.google.cloud.audit.AuthenticationInfo;
+import com.google.protobuf.Timestamp;
+import java.time.Clock;
+import java.time.Instant;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.HttpStatus;
@@ -41,14 +44,19 @@ public class LoggingController {
 
   private final LoggingClient loggingClient;
 
+  private final Clock clock;
+
   @GetMapping
   @ResponseStatus(value = HttpStatus.OK)
   void loggingShell(
       @RequestParam(value = TRACE_ID_PARAMETER_KEY) String traceId,
       @RequestAttribute(TokenInterceptor.INTERCEPTOR_USER_EMAIL_KEY) String userEmail)
       throws LogProcessingException {
+    Instant now = clock.instant();
     AuditLogRequest record =
         AuditLogRequest.newBuilder()
+            .setTimestamp(
+                Timestamp.newBuilder().setSeconds(now.getEpochSecond()).setNanos(now.getNano()))
             .setPayload(
                 AuditLog.newBuilder()
                     .setServiceName(SERVICE_NAME)
--- a/integration/testrunner/db.go
+++ b/integration/testrunner/db.go
@@ -66,13 +66,10 @@ func queryAuditLogs(ctx context.Context, tb testing.TB, query *bigquery.Query) [
 		if !ok {
 			tb.Fatalf("error converting query (%T) to string: %v", value[0], err)
 		}
-
+		tb.Logf("bq row is %s", value)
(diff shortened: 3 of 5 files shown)


clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
```diff
@@ -84,8 +85,9 @@ public AuditLogRequest process(AuditLogRequest auditLogRequest) throws LogProces
                   Instant.ofEpochSecond(
                       auditLogRequest.getTimestamp().getSeconds(),
                       auditLogRequest.getTimestamp().getNanos()))
-              .setResource(MonitoredResource.newBuilder(MONITORED_RESOURCE_TYPE).build())
+              // .setResource(MonitoredResource.newBuilder(MONITORED_RESOURCE_TYPE).build())
```
sethvargo, line 88: Is this intended?

capri-xiyue: I have not finished the change. I'm on the debugging process. No need to take a look at it now. But this change is intended. I found that our go code https://github.com/abcxyz/lumberjack/blob/554bab0378f22d2935cb8630e4c7e89f3fc8a908/clients/go/pkg/cloudlogging/processor.go#L128 doesn't override the `resource` field. Maybe we make both java and go consistent in setting the `resource` field. And based on our doc, this field can be auto generated by cloud logging. I don't see any value of overridden it by ourselves. Did I miss any context?

capri-xiyue: In addition, this part causes null fields in json maps which triggers incomplete unmarshal due to limitation in protojson. With the limitation of https://github.com/golang/protobuf/issues/1313, we won't be able to verify the timestamp is set in our e2e test and our test is also fragile. Therefore, I try to make the test logs without any null value in json maps or json arrays.

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
```diff
@@ -84,7 +83,6 @@ public AuditLogRequest process(AuditLogRequest auditLogRequest) throws LogProces
                   Instant.ofEpochSecond(
                       auditLogRequest.getTimestamp().getSeconds(),
                       auditLogRequest.getTimestamp().getNanos()))
-              .setResource(MonitoredResource.newBuilder(MONITORED_RESOURCE_TYPE).build())
```
capri-xiyue, line 87: This change is intended. I found that our go code https://github.com/abcxyz/lumberjack/blob/554bab0378f22d2935cb8630e4c7e89f3fc8a908/clients/go/pkg/cloudlogging/processor.go#L128 doesn't override the `resource` field. I think we need to make both java and go consistent in setting the `resource` field. And based on our doc, this field can be auto generated by cloud logging. I don't see any value of overridden it by ourselves.

capri-xiyue: In addition, this part causes null fields in json maps which triggers incomplete unmarshal due to limitation in protojson. With the limitation of https://github.com/golang/protobuf/issues/1313, we won't be able to verify the timestamp is set in our e2e test and our test is also fragile(fields whose order is after `resource` field can't be unmarshalled). Therefore, I try to make the test logs without any null value in json maps or json arrays. I still prefer not to changing the src code just to make test log consistent. But in this case, we have inconsistency in java and go src code when setting `resource` field, I think it makes sense to let java follow the go business logic which let cloud logging to auto generate the `resource` field.

integration/testrunner/db.go
```diff
@@ -66,13 +66,10 @@ func queryAuditLogs(ctx context.Context, tb testing.TB, query *bigquery.Query) [
 		if !ok {
 			tb.Fatalf("error converting query (%T) to string: %v", value[0], err)
 		}
-
+		tb.Logf("bq row is %s", value)
```
sqin2019, line 69: Is this intended or a residual of your debug?

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
```diff
@@ -23,7 +23,6 @@
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.cloud.MonitoredResource;
 import com.google.cloud.logging.LogEntry;
 import com.google.cloud.logging.Logging;
 import com.google.cloud.logging.LoggingException;
```
sqin2019, line 28: nit: may need to remove line `private static final String MONITORED_RESOURCE_TYPE = "global";` if no other usage of it.

integration/testrunner/db.go
capri-xiyue: Intended. During this debug, I found it is important to see what the bq row is becauase it could be different from the unmarshalled logentry. And this log will only appear when test fails.

sqin2019 (approved):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
yolocs: No. I thought cloud logging would fill this part. It's not critical to the audit logging.

clients/java-logger/shell/src/main/java/com/abcxyz/lumberjack/loggingshell/LoggingController.java
```diff
@@ -41,14 +44,19 @@ public class LoggingController {
...
   @ResponseStatus(value = HttpStatus.OK)
   void loggingShell(
       @RequestParam(value = TRACE_ID_PARAMETER_KEY) String traceId,
       @RequestAttribute(TokenInterceptor.INTERCEPTOR_USER_EMAIL_KEY) String userEmail)
       throws LogProcessingException {
+    Instant now = clock.instant();
     AuditLogRequest record =
         AuditLogRequest.newBuilder()
+            .setTimestamp(
+                Timestamp.newBuilder().setSeconds(now.getEpochSecond()).setNanos(now.getNano()))
```
yolocs, line 59: So cloud logging doesn't fill the timestamp automatically?

yolocs (approved):

clients/java-logger/shell/src/main/java/com/abcxyz/lumberjack/loggingshell/LoggingController.java
capri-xiyue: cloud logging set timestamp when no timestamp is provided
But in our code, we set the timestamp as null when no timestamp field is set in the auditlog request
Therefore, cloud logging won't set the timestamp for us
See code here
https://github.com/abcxyz/lumberjack/blob/554bab0378f22d2935cb8630e4c7e89f3fc8a908/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java#L85A

capri-xiyue: And see docs here `If this field is omitted in a new log entry, then Logging assigns it the current time.`
https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry. We didn't omit that field in our generated log entry

PR #384, opened 2023-02-16
fix: apply log mode option when initiating go client

sqin2019:
1. Updated java code since we want to treat LOG_MODE_UNSPECIFIED as FAIL_CLOSE as a safer default, to be consistent with go client. 
2. fixed go config validate and setDefault not called.
3. go client does not pick up the logmode, added via WithLogMode.


yolocs (approved):
Thank you!

PR #385, opened 2023-02-17
fix: omit timestamp when creating logentry when no timestamp is provided 

capri-xiyue:
omit timestamp when creating logentry when no timestamp is provided, cloud logging will automatically fill in the timestamp field.

### DIFF
--- a/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
+++ b/clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
@@ -31,7 +31,6 @@ import com.google.cloud.logging.Payload;
 import com.google.inject.Inject;
 import com.google.logging.v2.LogEntryOperation;
 import com.google.protobuf.InvalidProtocolBufferException;
-import com.google.protobuf.Timestamp;
 import com.google.protobuf.util.JsonFormat;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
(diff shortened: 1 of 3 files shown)


clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
```diff
@@ -77,12 +78,15 @@ public AuditLogRequest process(AuditLogRequest auditLogRequest) throws LogProces
...
                       .setLast(operation.getLast())
-                      .build())
-              .setTimestamp(
-                  Instant.ofEpochSecond(
-                      auditLogRequest.getTimestamp().getSeconds(),
-                      auditLogRequest.getTimestamp().getNanos()))
-              .build();
+                      .build());
+
+      if (!(auditLogRequest.getTimestamp()).equals(Timestamp.getDefaultInstance())) {
```
yolocs, line 83: Is the default instance null?

yolocs (approved):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/processor/CloudLoggingProcessor.java
sqin2019: shoud use hasTimestamp() instead?

capri-xiyue: default instance is not null. I tried to use null check but intellij indicates that timestamp will always not be null

capri-xiyue: > shoud use hasTimestamp() instead?

I didn't know we have such method in Java protobuf timestamp.  This is more clear. Updated it.

sqin2019 (approved):

PR #389, opened 2023-02-23
feat: enabled justification check in http

capri-xiyue:
part of #381 
1. enabled justification check in java http and golang http e2e
2. changed java jvs public key server in e2e to localhost


capri-xiyue:
I finished the Golang side change. Will continue to work on the Java side change. The integration test failed because of Java HTTP test cases.

clients/java-logger/shell/src/main/java/com/abcxyz/lumberjack/loggingshell/LoggingShellApplication.java
```diff
@@ -16,13 +16,45 @@
...
-  public static void main(String[] args) {
+  public static void main(String[] args) throws IOException {
+    HttpServer jwkServer = HttpServer.create(new InetSocketAddress(8081), 0);
+    jwkServer.createContext("/.well-known/jwks", new JWKHandler());
+    jwkServer.setExecutor(null); // creates a default executor
+    jwkServer.start();
     SpringApplication.run(LoggingShellApplication.class, args);
   }
+
+  static class JWKHandler implements HttpHandler {
```
sqin2019, line 39: Should we make this as util as well?

clients/java-logger/shell/src/test/java/com/abcxyz/lumberjack/loggingshell/LoggingControllerTest.java
```diff
@@ -44,15 +44,20 @@ public class LoggingControllerTest {
...
 
+  private static final String TEST_JVS_TOKEN =
+      "eyJhbGciOiJFUzI1NiIsImtpZCI6ImludGVnLWtleSIsInR5cCI6IkpXVCJ9.eyJhdWQiOlsiaHR0cCJdLCJleHAiOjE2NzcyODAzMjEsImlhdCI6MTY3NzI3NjcyMSwiaXNzIjoibHVtYmVyamFjay10ZXN0LXJ1bm5lciIsImp0aSI6IjkwMTdiZGJkLTViNDktNGVmOS1hZjllLTdmMDkxOWZmNGFiZiIsImp1c3RzIjpbeyJjYXRlZ29yeSI6InRlc3QiLCJ2YWx1ZSI6InRlc3QifV0sIm5iZiI6MTY3NzI3NjcyMSwic3ViIjoibHVtYmVyamFjay1pbnRlZyJ9.GzwrHJg2DbQAA64lOJcNGcBynFftIzY5eGUR9brHd9t6GTRImDHQUiuE4sQBYkDaKoEoPUh_cDca5UkGTbqmcg";
+
   private static final Timestamp TEST_TIMESTAMP =
       Timestamp.newBuilder().setSeconds(1577836800).build();
   private static final MockHttpServletRequestBuilder GET_REQUEST_BUILDER = get(REQUEST_PATH);
   private static final MockHttpServletRequestBuilder GET_REQUEST_BUILDER_WITH_EMAIL =
       get(REQUEST_PATH).requestAttr(TokenInterceptor.INTERCEPTOR_USER_EMAIL_KEY, TEST_EMAIL);
-  private static final MockHttpServletRequestBuilder GET_REQUEST_BUILDER_WITH_EMAIL_AND_TRACE_ID =
```
sqin2019, line 52: What happen if justification is missing in the header. should we only require token when justification is enabled?

clients/java-logger/shell/src/test/java/com/abcxyz/lumberjack/loggingshell/LoggingControllerTest.java
```diff
@@ -73,19 +78,21 @@ void loggingWithoutTraceIdCausesBadRequestError() throws Exception {
 
   @Test
   void loggingWithTraceIdReturnsOk() throws Exception {
```
sqin2019, line 80: Nit: we can assert all fields matching our test field in one test instead of multiple tests and each with one assertThat.

clients/java-logger/shell/src/main/java/com/abcxyz/lumberjack/loggingshell/LoggingShellApplication.java
capri-xiyue: I found java http and java grpc has two pom.xml. We can make this as a util but we can do it in separate PR.

clients/java-logger/shell/src/test/java/com/abcxyz/lumberjack/loggingshell/LoggingControllerTest.java
capri-xiyue: Yep. I also found other java code in lumberack has such multiple tests style. I will create a separate issue to fix it.

clients/java-logger/shell/src/main/resources/audit_logging.yml
```diff
@@ -23,6 +23,5 @@ backend:
     insecure_enabled: false
     impersonate_account:
 justification:
+  public_keys_endpoint: "http://0.0.0.0:8081/.well-known/jwks"
```
sethvargo, line 26: I think we still want `localhost` here?

clients/java-logger/shell/src/main/java/com/abcxyz/lumberjack/loggingshell/LoggingShellApplication.java
sqin2019: This belongs to jvs testing logics, we can add it in jvs. Not a blocker, can be an improvement later on.

clients/java-logger/shell/src/test/java/com/abcxyz/lumberjack/loggingshell/LoggingControllerTest.java
sqin2019: This could be a blocker for future local testing (if we plan to add) when we want to test disabled case?

capri-xiyue: Yes. If we want to test disabled case, we need to change it. But currently, I don't see any plans to test disabled case for now. In addition, we even removed requireJustification setting in GRPC e2e.

clients/java-logger/shell/src/main/resources/audit_logging.yml
capri-xiyue: done

clients/java-logger/shell/src/main/java/com/abcxyz/lumberjack/loggingshell/LoggingShellApplication.java
capri-xiyue: Will create a separate issue for it.

clients/java-logger/shell/src/test/java/com/abcxyz/lumberjack/loggingshell/LoggingControllerTest.java
capri-xiyue: https://github.com/abcxyz/lumberjack/issues/392

clients/java-logger/shell/src/main/java/com/abcxyz/lumberjack/loggingshell/LoggingShellApplication.java
capri-xiyue: https://github.com/abcxyz/lumberjack/issues/391

integration/testrunner/logging.go
```diff
@@ -27,6 +27,11 @@ func MakeAuditLogRequest(id, endpointURL string, requestTimeout time.Duration, a
 		return nil, fmt.Errorf("failed to create audit log http request: %w", err)
 	}
 
+	signedToken, err := justificationToken("talker-app")
```
sqin2019, line 30: This is used in http requests only, should be logging shell app instead? also wondering why we need this as audience, when verify we only check if justification is non-nil?

capri-xiyue: This is a typo. I changed it to `logging shell`. We didn't need this when validating justification. But previously grpc set the audience to `talker-app`. I don't want to change the existing behavior.

sqin2019 (approved):

clients/go/test/shell/config.yaml
```diff
@@ -0,0 +1,26 @@
+# Copyright 2022 Lumberjack authors (see AUTHORS file)
```
sqin2019, line 1: Oops missed this, nit: 2023

integration/testrunner/logging.go
sqin2019: Ok, another way is to use a dummy or no audience at all so that we don't need to pass any thing that we don't need.

clients/java-logger/grpc-test-app/src/main/java/abcxyz/lumberjack/test/talker/TalkerService.java
```diff
@@ -134,7 +135,8 @@ public void handle(HttpExchange t) throws IOException {
 
   /** Main launches the server from the command line. */
   public static void main(String[] args) throws IOException, InterruptedException {
-    HttpServer jwkServer = HttpServer.create(new InetSocketAddress(8081), 0);
+    HttpServer jwkServer =
+        HttpServer.create(new InetSocketAddress(InetAddress.getLocalHost(), 8080), 0);
```
yolocs, line 139: Isn't 8080 used by the TalkerService itself?

clients/java-logger/shell/src/main/java/com/abcxyz/lumberjack/loggingshell/LoggingShellApplication.java
```diff
@@ -16,13 +16,47 @@
...
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
 /** Entry point for the Logging Shell/Test app. */
 @SpringBootApplication
 public class LoggingShellApplication {
-  public static void main(String[] args) {
+  public static void main(String[] args) throws IOException {
+    HttpServer jwkServer =
+        HttpServer.create(new InetSocketAddress(InetAddress.getLocalHost(), 8080), 0);
```
yolocs, line 34: Same question

integration/testrunner/db.go
```diff
@@ -173,10 +174,21 @@ func diffLogEntry(tb testing.TB, logEntry *loggingpb.LogEntry, requireJustificat
...
+	tb.Helper()
+	justification, ok := jsonPayloadInfo.Metadata.AsMap()["justification"]
+	if !ok {
+		tb.Fatalf("queryResult field %v doesn't exist", "jsonPayload.metadata.justification")
+	}
+	b, err := json.Marshal(justification)
+	if err != nil {
+		tb.Fatalf("failed to marshal justification: %v", err)
+	}
+	if string(b) == "null" {
```
yolocs, line 190: Should we also need to check against empty string?

yolocs (approved):

clients/java-logger/grpc-test-app/src/main/java/abcxyz/lumberjack/test/talker/TalkerService.java
capri-xiyue: I found in Golang, the grpc app and shell app expose 8080, but in Java I didn't find it. In addtional, in java, we have to specify the path `/.well-known/jwks` in the public key endpoint.

clients/java-logger/shell/src/main/java/com/abcxyz/lumberjack/loggingshell/LoggingShellApplication.java
capri-xiyue: Same above

integration/testrunner/db.go
capri-xiyue: I checked bq, looks like the default value is null. But I can't find official docs for it. To make it more robust, I will add empty string check here.

PR #390, opened 2023-02-27
fix: remove some jsonProperty, and rename those variables to unify with go version.

sailorlqh:


clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/config/AuditLoggingConfiguration.java
```diff
@@ -40,22 +40,20 @@ public class AuditLoggingConfiguration {
   @JsonProperty("log_mode")
```
sqin2019, line 40: This should be removed if we use global naming strategy with object mapper.

Issue #391, opened 2023-02-27
clean duplicate jvs public key server in java grpc and java shell


capri-xiyue:
### TL;DR

Duplicate code in https://github.com/abcxyz/lumberjack/pull/389/files/74efe1b4688af70ab74622de626308dc3268d4c1#diff-bbe752cb9db5b89eae789fdd0770ba36c69177405618b9a3182d3bbeb586290fR31 and https://github.com/abcxyz/lumberjack/blob/a0ec01cf9483c17ae1fec686064036817f8b0be8/clients/java-logger/grpc-test-app/src/main/java/abcxyz/lumberjack/test/talker/TalkerService.java#L113

PR #404, opened 2023-03-17
fix: add utils for handling jvs public key, and remove duplicate keys

sailorlqh:
fix issue: #391


yolocs:
I don't think we should add this test key to non-test code. Anything under `clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/` is non-test code.

Plus, the public / private key used in integration test are shared across Go and Java. I think the proper fix should be:

1) Put the keys in files under `integration` folder
2) Update the test app dockerfiles to copy the public key into the container image
3) Update the test apps to use read the public key file
4) Update integration test to accept the test private key as a config.

cc: @capri-xiyue

capri-xiyue:
Agree that we should not add this test key to non-test code

capri-xiyue:
> I don't think we should add this test key to non-test code. Anything under `clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/` is non-test code.
> 
> Plus, the public / private key used in integration test are shared across Go and Java. I think the proper fix should be:
> 
> 1. Put the keys in files under `integration` folder
> 2. Update the test app dockerfiles to copy the public key into the container image
> 3. Update the test apps to use read the public key file
> 4. Update integration test to accept the test private key as a config.
> 
> cc: @capri-xiyue

Yes, we should not add this test key to non-test code

sqin2019:
@yolocs As an improvement, should we put this JWKHandler in JVS pkg as a test util?

PR #409, opened 2023-03-24
feat: introduce new log types

yolocs:
Fix: https://github.com/abcxyz/lumberjack/issues/407

### DIFF
--- a/protos/v1alpha1/audit_log_request.proto
+++ b/protos/v1alpha1/audit_log_request.proto
@@ -49,11 +49,11 @@ message AuditLogRequest {
     // APIs.
     DATA_ACCESS = 2 [(log_name) = "audit.abcxyz/data_access"];
 
-    // For audit logging system (vs. human) generated events.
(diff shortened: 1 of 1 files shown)


verbanicm (approved):

protos/v1alpha1/audit_log_request.proto
```diff
@@ -48,6 +48,12 @@ message AuditLogRequest {
     // Reads of configuration data and all access to user data through public
     // APIs.
     DATA_ACCESS = 2 [(log_name) = "audit.abcxyz/data_access"];
+
+    // For audit logging system (vs. human) generated events.
+    SYSTEM_EVENT = 3 [(log_name) = "audit.abcxyz/system_event"];
```
sethvargo, line 53: Philosophical - should we leave some room between system_event and the other kinds? I would image we might want system_event to be 50 or something.

sethvargo (approved):

protos/v1alpha1/audit_log_request.proto
yolocs: Good idea.

Issue #427, opened 2023-07-31
Feat: add a validate cli tool to validate lumberjack log


sqin2019:

sqin2019:
follow up: #433

sqin2019:
pending - add a new release

sqin2019:
followup #439

PR #431, opened 2023-08-03
feat: lumberjack log validation cli

sqin2019:
part of #427


pkg/cli/root.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+
+	"github.com/abcxyz/lumberjack/internal/version"
+	"github.com/abcxyz/pkg/cli"
+)
+
+// rootCmd defines the starting command structure.
+var rootCmd = func() cli.Command {
+	return &cli.RootCommand{
+		Name:    "ljctl",
```
sethvargo, line 28: Hmm - we should probably name this `lumberjackctl`. `ljctl` is kinda weird.

pkg/cli/validate.go
```diff
@@ -0,0 +1,130 @@
...
+		Aliases: []string{"l"},
+		Target:  &c.flagLog,
+		Example: `{"foo":"bar"}`,
+		Usage: `The lumberjack/data access log, in JSON format. Set the value to` +
+			` "-" to read from stdin, it stops reading when it reaches end of file`,
+	})
+
+	f.BoolVar(&cli.BoolVar{
+		Name:    "additional-check",
+		Aliases: []string{"ac"},
```
sethvargo, line 74: I don't think we should have an alias for this one.

pkg/cli/validate.go
```diff
@@ -0,0 +1,130 @@
...
+	if c.flagLog == "" {
+		return fmt.Errorf("log is required")
+	}
+
+	if c.flagLog == "-" {
+		// Read log from stdin
+		log, err := c.readFromStdin(ctx, "Enter log: ")
+		if err != nil {
+			return fmt.Errorf("failed to get log from prompt: %w", err)
+		}
```
sethvargo, line 102: We can actually get rid of this now and call `c.Prompt()` 😄

sqin2019: Chen and I [discussed about this](https://github.com/abcxyz/lumberjack/pull/428#discussion_r1281193787), the existing c.Prompt() only reads one line as it uses bufio scanner.scan, whereas lumberjack logs are multi-lines. I am thinking I can add a multi-line reader here and then move it to pkg. Any suggestions?

pkg/cli/root.go
sqin2019: Cool, I am bad at naming, was thinking maybe shorter is preferred.

yolocs: I vote for "lumberctl" :) After all, we are looking at logs (lumber).

pkg/cli/validate.go
sethvargo: Ah good point. Then this is probably fine.

yolocs: _Could_ be multiple lines. I'm thinking `cat log.json | cli validate -log -` style

pkg/cli/root.go
sqin2019: Changed to the shorter version lumberctl, let me know if any concerns. Thanks both!

yolocs (approved):

pkg/cli/root.go
sqin2019: @sethvargo friendly ping

sethvargo (approved):

PR #434, opened 2023-08-07
feat: query cloud logging logs

sqin2019:
upcoming PR to add cli that pulls and validates logs. part of #427


pkg/cloudlogging/puller.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	if p.retry == nil {
+		p.retry = retry.WithMaxRetries(5, retry.NewFibonacci(500*time.Millisecond))
+	}
+	return p
+}
+
+// Pull pulls a list of log entries given request.
+func (p *Puller) Pull(ctx context.Context, req *loggingpb.ListLogEntriesRequest) ([]*loggingpb.LogEntry, error) {
```
yolocs, line 66: I wonder if we should have a `DefaultPull` with a preset `req`. The `req` will be based on log name.

sqin2019: Why? I am thinking this can live in the cli command?

yolocs: I'm fine with that as well

yolocs: Actually I changed my mind - I think it's better if we make `filter` an input here. A `Puller` should be scoped to a project/folder/org on initialization.

The `Pull` here is essential "pull once and see what you got" operation. This is what we immediately need. I'm thinking maybe later we can have a better approach to "tail the logs and see if any matching logs coming up".

sqin2019: I am not sure I follow, I am sharing my [plan](https://github.com/abcxyz/lumberjack/blob/sqin/cli/pkg/cli/pull_validate.go#L191) here - we can form the request in the cli however we want. For scoping the Puller, there is only one cloud logging client though. We can have a quick sync if you want.

sqin2019: ah..I got that scoping the Puller part, but I am wondering why we should do that...

pkg/cloudlogging/puller.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		return nil
+	}); err != nil {
+		return nil, fmt.Errorf("failed to list log entries: %w", err)
+	}
+	return ls, nil
+}
+
+// SteamPull pulls live log entries, it stops pulling after logNum of log entries.
+func (p *Puller) SteamPull(ctx context.Context, filter string, logNum int) ([]*loggingpb.LogEntry, error) {
```
sqin2019, line 122: @yolocs to communicate the idea (haven't updated test yet), are you suggesting to have both Pull and StreamPull?

pkg/cloudlogging/puller.go
```diff
@@ -0,0 +1,119 @@
...
+		p = opt(p)
+	}
+
+	if p.retry == nil {
+		p.retry = retry.WithMaxRetries(5, retry.NewFibonacci(500*time.Millisecond))
+	}
+	return p
+}
+
+// Pull pulls up to maxCount of log entries given log filter.
```
yolocs, line 85:
```suggestion
// Pull pulls up to maxCount of log entries for the given log filter.
```

yolocs (approved):

pkg/cloudlogging/puller.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/sethvargo/go-retry"
+	"google.golang.org/api/iterator"
+
+	logging "cloud.google.com/go/logging/apiv2"
+)
+
+// LoggingClient interface for pulling log entries.
+type LoggingClient interface {
+	ListLogEntries(context.Context, *loggingpb.ListLogEntriesRequest, ...gax.CallOption) *logging.LogEntryIterator
+}
```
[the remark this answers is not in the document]
sqin2019: probably not, since it still kinda coupled with cloud logging api, removing.

PR #435, opened 2023-08-09
feat: pull and validate cli

sqin2019:


pkg/cli/validate_pull.go
```diff
@@ -0,0 +1,233 @@
...
+
+// logPuller interface that pulls log entries from cloud logging.
+type logPuller interface {
+	Pull(context.Context, string, int) ([]*loggingpb.LogEntry, error)
+}
+
+var _ cli.Command = (*ValidatePullCommand)(nil)
+
+// ValidatePullCommand validates lumberjack logs pulled.
+type ValidatePullCommand struct {
```
sqin2019, line 47: Can't think of a better name...

pkg/cli/root.go
```diff
@@ -31,6 +31,9 @@ var rootCmd = func() cli.Command {
 			"validate": func() cli.Command {
 				return &ValidateCommand{}
 			},
+			"pull": func() cli.Command {
+				return &ValidatePullCommand{}
```
sethvargo, line 35: Should this just be `PullCommand`?

I wonder if "tail" might be a better name (referring to the Linux command `tail`).

sqin2019: It does both things pull and validate, I am ok with `PullCommand` since lumberctl is a validation tool, it should imply it will also run validation. 

If we should use `tail`, it'd be better to rename pull everywhere? What will be a good replacement for `LogPuller`?

sethvargo: Maybe tail is a bad name then :)

sqin2019: Is `LogTailer` a good name? I am not native english speaker, can't tell...

yolocs (commented):
As we've discussed offline, let's name the subcommand `tail` and add `-validate` flag to enable validation.

pkg/cli/root_test.go
```diff
@@ -25,7 +25,7 @@ func TestRootCommand_Help(t *testing.T) {
 	exp := `
 Usage: lumberctl COMMAND
 
-  pull        Pulls and Validates lumberjack logs from Cloud logging
+  tail        Tail lumberjack logs from Cloud logging and validate them when validation enabled
```
yolocs, line 28: I think it's better we mention validation in the flag
```suggestion
  tail        Tail lumberjack logs from GCP cloud logging
```

pkg/cli/tail.go
```diff
@@ -0,0 +1,254 @@
...
+
+	f.BoolVar(&cli.BoolVar{
+		Name:    "remove-lumberjack-log-type",
+		Target:  &c.flagRemoveLumberjackLogType,
+		Default: false,
+		Usage:   `Turn on to remove lumberjack log type log filter`,
+	})
+
+	f.BoolVar(&cli.BoolVar{
+		Name:    "validate-with-additional-check",
```
yolocs, line 151: I think this should be `additional-check` and is only useful when `-validate` is on

pkg/cli/tail.go
```diff
@@ -0,0 +1,254 @@
...
+
+	ls, err := p.Pull(ctx, c.getFilter(), c.flagMaxNum)
+	if err != nil {
+		return nil, fmt.Errorf("failed to pull logs: %w", err)
+	}
+
+	return ls, nil
+}
+
+func (c *TailCommand) getFilter() string {
```
yolocs, line 240: Go rarely uses `get`.  Use something like `queryFilter()` instead

pkg/cli/tail.go
```diff
@@ -0,0 +1,254 @@
...
+		p = c.testPuller
+	} else {
+		logClient, err := logging.NewClient(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create logging client: %w", err)
+		}
+		p = cloudlogging.NewPuller(ctx, logClient, c.flagResource)
+	}
+
+	ls, err := p.Pull(ctx, c.getFilter(), c.flagMaxNum)
```
yolocs, line 232: Should debug output the filter.

pkg/cli/tail.go
```diff
@@ -0,0 +1,254 @@
...
+		Name:    "duration",
+		Aliases: []string{"d"},
+		Target:  &c.flagDuration,
+		Example: "2h",
+		Default: 24 * time.Hour,
+		Usage:   `How far back to search for log entries, default is 24 hours`,
+	})
+
+	f.StringVar(&cli.StringVar{
+		Name:    "query",
```
yolocs, line 136: This is a bit misleading. This is really _additional_ filters we append to the query, right? Should it be `additional-filter` and make it clear in the usage doc

yolocs (changes requested):

pkg/cli/tail.go
```diff
@@ -0,0 +1,254 @@
...
+		Usage: `Optional custom log queries, see more on ` +
+			`https://cloud.google.com/logging/docs/view/logging-query-language`,
+	})
+
+	f.BoolVar(&cli.BoolVar{
+		Name:    "remove-lumberjack-log-type",
+		Target:  &c.flagRemoveLumberjackLogType,
+		Default: false,
+		Usage:   `Turn on to remove lumberjack log type log filter`,
+	})
```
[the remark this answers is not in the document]
sqin2019: I found it useful when I am testing the cli and I don't have lumberjack logs in order to see the output. It might also be helpful when users forget to add the log type and want to query specific test logs?

pkg/cli/tail.go
```diff
@@ -0,0 +1,254 @@
...
+		if err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("failed to marshal log to json (InsertId: %q): %w", l.InsertId, err))
+			continue
+		}
+
+		// Output log entry in JSON format if validation is not enabled.
+		if !c.flagValidate && !c.flagValidateWithAdditionalCheck {
+			c.Outf(string(js))
+			continue
+		}
```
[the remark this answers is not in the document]
sqin2019: Since the log is long, it will not give a clear view for validation results.

pkg/cli/tail.go
sqin2019: I wonder do we really need to output filter, it should be obvious what filters are used (I can add the lumberjack log type to the cmd description)?

pkg/cli/tail.go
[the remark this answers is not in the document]
sqin2019: update to use override-filter flag, let me know what you think!

pkg/cli/tail.go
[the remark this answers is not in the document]
sqin2019: updated.

pkg/cli/root.go
sqin2019: Switched to tail, but kept logPuller. Let me know if it is confusing. Thanks!

pkg/cli/tail.go
```diff
@@ -0,0 +1,257 @@
...
+	return `Tail lumberjack logs from GCP Cloud logging`
+}
+
+func (c *TailCommand) Help() string {
+	return `
+Usage: {{ COMMAND }} [options]
+
+Tails and validates the latest lumberjack log in the last 2 hours in the scope:
+
+      {{ COMMAND }} -scope "project/foo" -validate
```
yolocs, line 78: Should it be `projects/foo`?

pkg/cli/tail.go
```diff
@@ -0,0 +1,257 @@
...
+
+      {{ COMMAND }} -scope "project/foo" -additional-filter "resource.type = \"foo\""
+
+Tails and validates (with additional check) the latest 10 lumberjack log in the last 4 hours in the scope:
+
+      {{ COMMAND }} -scope "project/foo" -max-num 10 -duration 4h -validate -additional-check
+
+Tails and validates the latest log using override log filter only:
+
+      {{ COMMAND }} -scope "project/foo" -override-filter "YOUR_FILTER"
```
yolocs, line 90: Don't provide example for hidden flags

pkg/cli/tail.go
```diff
@@ -0,0 +1,257 @@
...
+			continue
+		}
+
+		// Output tailed log.
+		c.Outf(string(js))
+
+		// Output validation result if validation is enabled.
+		if c.flagValidate {
+			if err := validation.Validate(string(js), extra...); err != nil {
+				retErr = errors.Join(retErr, fmt.Errorf("failed to validate log (InsertId: %q): %w", l.InsertId, err))
```
yolocs, line 213: We should output error right away `c.Errf` so that the error message will be right after the log itself. Otherwise it would be hard to map the error message back to the log.

pkg/cli/tail.go
```diff
@@ -0,0 +1,257 @@
...
+			} else {
+				c.Outf("Successfully validated log (InsertId: %q)", l.InsertId)
+			}
+		}
+		c.Outf("\n")
+	}
+
+	return retErr
```
yolocs, line 221: What you can do here is the count the number of failed logs. If it's great than 0, return `fmt.Errorf("validation failed for %d logs (out of %d)")`

sqin2019: Updated.

pkg/cli/tail_test.go
```diff
@@ -0,0 +1,235 @@
...
+							"principal_email": structpb.NewStringValue("foo@bet.com"),
+						},
+					}),
+				},
+			},
+		},
+		Labels: map[string]string{"environment": "dev", "accessing_process_name": "foo_apn"},
+	}
+
+	validLogJson, err := protojson.Marshal(validLog)
```
sqin2019, line 55: Two issues with protojson.Marshal:
1. We use protojson.Marshal to marshal the log entry tailed to JSON in tail.go as well, it seems protojson.Marshal does not produce deterministic results.  
2. I always got `failed to marshal log to json (InsertId: "-oojsuhe1sciu"): proto: google.protobuf.Any: unable to resolve "type.googleapis.com/google.iam.v1.logging.AuditData": not found` error when I try to parse a log entry with proto payload. Wondering why and if it is a risk (though we only expect json payload).

sqin2019: update:
1. it is known issue - https://github.com/golang/protobuf/issues/1121, going to stripe the spaces in unit test as a workaround.

pkg/cli/tail_test.go
```diff
@@ -52,10 +52,11 @@ func TestTailCommand(t *testing.T) {
 		Labels: map[string]string{"environment": "dev", "accessing_process_name": "foo_apn"},
 	}
 
-	validLogJSON, err := protojson.Marshal(validLog)
+	bs, err := protojson.Marshal(validLog)
```
yolocs, line 55: I missed it earlier. In lines 202-207 - for string comparison (or other primitive types), use `!=` instead of `cmp.Diff`

pkg/cli/tail.go
```diff
@@ -200,8 +201,10 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
 			continue
 		}
 
-		// Output tailed log.
-		c.Outf(string(js))
+		// Output tailed log, all spaces are stripped to reduce unit test flakiness
+		// as protojson.Marshal can produce inconsistent output. See issue
+		// https://github.com/golang/protobuf/issues/1121.
+		c.Outf(strings.Replace(string(js), " ", "", -1))
```
yolocs, line 207: Write a helper func `stripSpaces` and use it in test as well

yolocs (changes requested):

pkg/cli/tail_test.go
sqin2019: update:
2. We should be good as we only expecting json payload.

yolocs: Yes, we don't expect proto payload

yolocs (approved):

pkg/cli/tail.go
sqin2019: Resolving, feel free to reopen.

pkg/cli/tail.go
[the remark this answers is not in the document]
sqin2019: Resolving, feel free to reopen

pkg/cli/root.go
sqin2019: Resolving, feel free to reopen

PR #436, opened 2023-08-11
feat: add binary release for lumberctl

sqin2019:
part of #427

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -47,6 +47,28 @@ builds:
       - 'amd64'
       - 'arm64'
 
+  -
+    id: lumberctl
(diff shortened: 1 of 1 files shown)


sethvargo (approved):

.goreleaser.yaml
```diff
@@ -47,6 +47,28 @@ builds:
       - 'amd64'
       - 'arm64'
 
+  -
+    id: lumberctl
```
sethvargo, line 51: Quote your YAML plz

Also, do we have `yamllint` installed on this repo?

sqin2019: Don't think so, I will open a separate PR to add that, thx!

PR #437, opened 2023-08-11
chore: add yaml lint

sqin2019:


sqin2019:
maybe we should get rid of third party pkgs, for now I am going to ignore yaml lint errors from them

sethvargo:
I think it might support that, but we don't have a native option for it. I'd have to dig in and look at the docs, but I don't have time at the moment. If someone else wants to look into it, that'd be great.

sqin2019:
> I think it might support that, but we don't have a native option for it. I'd have to dig in and look at the docs, but I don't have time at the moment. If someone else wants to look into it, that'd be great.

Chen is working on removing the third party pkgs to unblock. If not I can take a look on how to add ignore paths to pkg, ignore paths can be supported using yamllint config - https://yamllint.readthedocs.io/en/stable/configuration.html#ignoring-paths

sethvargo:
> Chen is working on removing the third party pkgs to unblock. If not I can take a look on how to add ignore paths to pkg, ignore paths can be supported using yamllint config - https://yamllint.readthedocs.io/en/stable/configuration.html#ignoring-paths

Ah then we should be good. We just need to feed that configuration through from a configuration variable in the yaml-lint in pkg.

PR #440, opened 2023-08-14
chore: add yamllint

sailorlqh:
Forked from https://github.com/abcxyz/lumberjack/pull/437. 

Added `.yamllint.yml` file, this file is the same as that from `abcxyz/pkg/.yamllint.yml` only added the `ignore` rule to ignore lint from third_party code.

Thanks @sqin2019 for contributing and researching on the solutions.


.yamllint.yml
```diff
@@ -0,0 +1,100 @@
+# Copyright 2023 The Authors (see AUTHORS file)
```
sqin2019, line 1: I was thinking instead of using a entire .yamllint.yaml , we could use an optional workflow input variable like ignore path [here](https://github.com/abcxyz/pkg/blob/main/.github/workflows/yaml-lint.yml#L28).

sailorlqh: `ignore` can't be used as a flag, if we want make ignore as a flag in the cmd, the option is using `-d {ignore: /third_party/*}`. But `-d` and `-c` can't be used at the same time. See: https://github.com/abcxyz/lumberjack/actions/runs/5860727366/job/15889390357.

So I think having a `.yamllint.yml`, and `pkg` will use that yaml file to do lint by default should be the better option.

sailorlqh: We can also update`yamllint.yml` in abcxyz/pkg (https://github.com/abcxyz/pkg/blob/main/.yamllint.yml) to ignore any code in `third_party` directory if @sethvargo and @yolocs think this is a common pattern that can apply to more projects.

sailorlqh:
Removed the yamllint now, as https://github.com/abcxyz/lumberjack/pull/441 solved the problem.

.yamllint.yml
sailorlqh: Removed the `yamllint.yml`, as https://github.com/abcxyz/lumberjack/pull/441 solved the problem of lint from third_party packages.

Though I think it might worth adding `ignore` to `abcxyz/pkg/.yamllint.yml`

yolocs (approved):

sqin2019 (approved):

sethvargo:
It makes sense to add `thirdparty`, `third_party`, and `vendor` to yamllint in pkg.

sethvargo (approved):
LGTM - the latest version of yamllint in pkg also ignores common vendor directories including `vendor`, `third_party`, and `thirdparty`. However, I agree that cleaning up the third_party directory is actually a much better solution for lumberjack 😄 .

.yamllint.yml
sethvargo: Done

PR #446, opened 2023-08-17
feat: add stream puller

sailorlqh:


pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,45 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
 	}
 	return ls, nil
 }
+
+// SteamPull pulls live log entries, it stops pulling after logNum of log entries.
+func (p *Puller) SteamPull(ctx context.Context, filter string, logNum int) ([]*loggingpb.LogEntry, error) {
```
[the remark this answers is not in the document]
sqin2019: In that case, we don't need to return []*loggingpb.LogEntry.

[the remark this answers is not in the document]
sailorlqh: Changed to use channel

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,47 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+// SteamPull pulls live log entries, it stops pulling after logNum of log entries.
+func (p *Puller) SteamPull(ctx context.Context, filter string, logNum int, result chan<- []*loggingpb.LogEntry) error {
+	logger := logger.FromContext(ctx)
+	if err := retry.Do(ctx, p.retry, func(ctx context.Context) error {
+		stream, err := p.client.TailLogEntries(ctx)
+		if err != nil {
+			return retry.RetryableError(fmt.Errorf("failed to create stream: %w", err))
+		}
+		defer func() {
+			if err := stream.CloseSend(); err != nil {
```
sqin2019, line 126: Let's add a member field in the Puller struct to include this stream client so that here we don't need to deal with closing the client.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,47 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+		}()
+		req := &loggingpb.TailLogEntriesRequest{
+			ResourceNames: []string{p.resource},
+			Filter:        filter,
+		}
+		if err := stream.Send(req); err != nil {
+			return retry.RetryableError(fmt.Errorf("failed to send request: %w", err))
+		}
+
+		for counter := 0; counter < logNum; {
```
sqin2019, line 138: Forgot to mention one of the feedback I got from Seth and Chen is to simulate the behaviour of linux [tail -f ](https://www.geeksforgeeks.org/tail-command-linux-examples/), let's get rid of logNum, and abort when interrupted.

pkg/cloudlogging/puller_test.go
```diff
@@ -98,7 +98,69 @@ func TestPull(t *testing.T) {
...
+			ctx := context.Background()
+			fakeClient := setupFakeClient(t, ctx, tc.server)
+			p := NewPuller(
+				ctx,
+				fakeClient,
+				testResource,
+				WithRetry(retry.WithMaxRetries(0, retry.NewFibonacci(500*time.Millisecond))),
+			)
+			ch := make(chan []*loggingpb.LogEntry, 1)
+			gotErr := p.SteamPull(ctx, tc.filter, 1, ch)
```
sqin2019, line 154: Close the ch after it is done? close(ch)

pkg/cloudlogging/puller_test.go
```diff
@@ -98,7 +98,69 @@ func TestPull(t *testing.T) {
...
+			if diff := testutil.DiffErrString(gotErr, tc.wantErrSubstr); diff != "" {
+				t.Errorf("Process(%+v) got unexpected error substring: %v", tc.name, diff)
+			}
+
+			if diff := cmp.Diff(tc.wantResult, <-ch, protocmp.Transform()); diff != "" {
+				t.Errorf("Process(%+v) got result diff (-want, +got): %v", tc.name, diff)
+			}
+
+			if diff := cmp.Diff(tc.wantReq, tc.server.tailReq, protocmp.Transform()); diff != "" {
```
sqin2019, line 163: All wantReq is nil?

sailorlqh: The API of TailLogEntries is `TailLogEntries(server loggingpb.LoggingServiceV2_TailLogEntriesServer) error`, therefore, we can't just store req the way we did for `ListLogEntries`.

I will remove the `diff wantReq` part

pkg/cloudlogging/puller_test.go
```diff
@@ -98,7 +98,69 @@ func TestPull(t *testing.T) {
...
+	cases := []struct {
+		name          string
+		server        *fakeServer
+		filter        string
+		wantReq       *loggingpb.TailLogEntriesRequest
+		wantResult    []*loggingpb.LogEntry
+		wantErrSubstr string
+	}{
+		{
+			name:   "success",
```
yolocs, line 120: Testing with channel tends to be trickier. You need to 1) check if the channel has received the expected number of logs 2) have timeout because if the code has bug that doesn't return the logs, the test will stuck. A rough example:

```go
var gotLogs []*LogEntry
ch := make(chan *LogEntry)
done := make(chan struct{}, 1)
t.Cleanup(func() {
  close(ch)
  close(done)
})

ctx, cancel := context.WithCancel(context.Background())

// Start a go routine to accept logs
go func() {
  for l := range ch {
    gotLogs = append(gotLogs, l)
    if len(gotLogs) == wantNum {
      done <- struct{}{}
      break;
    }
  }
}

if err := StreamPull(ctx, filter, ch); err != nil {
  // Handle err
}



select {
case <-done:
  // got enough logs
case <-time.After(2*time.Second):
  // timeout
}

// Either way, cancel the context to stop the pull
cancel()

// Compare gotLogs with wantLogs
```

pkg/cloudlogging/puller_test.go
sqin2019: Maybe you can try creating a fake server that implements LoggingServiceV2_TailLogEntriesServer
```
type tailLogEntriesServer struct {
	grpc.ServerStream
	gotReq *loggingpb.TailLogEntriesRequest
	resp   *loggingpb.TailLogEntriesResponse
}

func (x *tailLogEntriesServer) Send(m *loggingpb.TailLogEntriesResponse) error {
	x.resp = m
	return nil
}

func (x *tailLogEntriesServer) Recv() (*loggingpb.TailLogEntriesRequest, error) {
	return x.gotReq, nil
}
```

yolocs (changes requested):

pkg/cloudlogging/puller_test.go
sqin2019: I wonder if we can simplify here, just assume there is only log and close the channel right after to signal exit.

yolocs: Could you elaborate? Not sure I follow

sqin2019: sth like, close(gotLog) should exit the function.
```
			gotLog := make(chan *loggingpb.LogEntry, 1)
			gotErr := p.StreamPull(ctx, filter, gotLog)
			close(gotLog)
```

yolocs: It's possible that when you `close(gotLog)` the logs haven't entered the channel yet.

sqin2019: Even on the same goroutine?

pkg/cloudlogging/puller.go
sailorlqh: Done.

pkg/cloudlogging/puller.go
sailorlqh: Remove logNum

pkg/cloudlogging/puller_test.go
sailorlqh: I think from the official, channel should be closed by sender instead of receiver, so I think it should be closed by `streampull()`

https://go.dev/tour/concurrency/4

pkg/cloudlogging/puller_test.go
sailorlqh: Used another way, seems to be easier.

pkg/cloudlogging/puller_test.go
sailorlqh: Updated.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,47 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+			}
+			if err != nil {
+				return retry.RetryableError(fmt.Errorf("failed to receive response: %w", err))
+			}
+			if resp.Entries != nil {
+				counter += len(resp.Entries)
+				result <- resp.GetEntries()
+			}
+		}
+		close(result)
```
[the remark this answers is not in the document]
sailorlqh: From this link(https://go.dev/tour/concurrency/4), I think we should have sender close the channel. However, from the same link, I think we can also don't close any channel, and just leave it there.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,47 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+			resp, err := stream.Recv()
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			if err != nil {
+				return retry.RetryableError(fmt.Errorf("failed to receive response: %w", err))
+			}
+			if resp.Entries != nil {
+				counter += len(resp.Entries)
+				result <- resp.GetEntries()
```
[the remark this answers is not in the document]
sailorlqh: Done.

pkg/cloudlogging/puller.go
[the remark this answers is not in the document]
sailorlqh: Done.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,47 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
 	}
 	return ls, nil
 }
+
+// SteamPull pulls live log entries, it stops pulling after logNum of log entries.
+func (p *Puller) SteamPull(ctx context.Context, filter string, logNum int, result chan<- []*loggingpb.LogEntry) error {
```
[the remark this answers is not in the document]
sailorlqh: Updated.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,47 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+		}
+		if err := stream.Send(req); err != nil {
+			return retry.RetryableError(fmt.Errorf("failed to send request: %w", err))
+		}
+
+		for counter := 0; counter < logNum; {
+			resp, err := stream.Recv()
+			if errors.Is(err, io.EOF) {
+				break
+			}
```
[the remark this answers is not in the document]
sailorlqh: `io.EOF` here is for starting a new retry so StreamPull can keep pulling when there are new logs. ctrl+c won't affect this.

pkg/cloudlogging/puller_test.go
```diff
@@ -98,7 +98,109 @@ func TestPull(t *testing.T) {
...
+			)
+
+			go p.StreamPull(ctx, tc.filter, ch, errCh, streamClient)
+
+			select {
+			case <-done:
+				err := streamClient.CloseSend()
+				if err != nil {
+					t.Fatalf("failed to close streamClient %v", err)
+				}
```
yolocs, line 190: We should cancel context? That's what will happen if a user hit ctrl+c.

yolocs (changes requested):

pkg/cloudlogging/puller_test.go
sailorlqh: Use `context.WithCancel` now.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,38 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+					logCh <- v
+				}
+			}
+		}
+	}); err != nil {
+		errCh <- err
+		// Sender should close the channel. If the caller close the channel, it
+		// could lead to sending on a closed channel and cause a panic
+		close(errCh)
+		close(logCh)
```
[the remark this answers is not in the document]
sailorlqh: Updated

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,38 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
 	}
 	return ls, nil
 }
+
+// SteamPull pulls live log entries, it won't stop until a cancel signal happens.
+func (p *Puller) StreamPull(ctx context.Context, filter string, logCh chan<- *loggingpb.LogEntry, errCh chan<- error, tailLogEntriesClient loggingpb.LoggingServiceV2_TailLogEntriesClient) {
```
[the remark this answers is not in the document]
sailorlqh: Updated.

pkg/cloudlogging/puller.go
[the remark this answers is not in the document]
yolocs: I was wrong. This wouldn't work because this func will be blocking.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,34 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
 	}
 	return ls, nil
 }
+
+// SteamPull pulls live log entries, it won't stop until a cancel signal happens.
+func (p *Puller) StreamPull(ctx context.Context, filter string, logCh chan<- *loggingpb.LogEntry, tailLogEntriesClient loggingpb.LoggingServiceV2_TailLogEntriesClient) (chan<- *loggingpb.LogEntry, error) {
```
yolocs: The func shouldn't accept the stream client. It should be invisible to the caller.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,34 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+			if resp.GetEntries() != nil {
+				for _, v := range resp.GetEntries() {
+					logCh <- v
+				}
+			}
+		}
+	}); err != nil {
+		return logCh, fmt.Errorf("failed to pull log entries: %w", err)
+	}
+	return logCh, nil
```
yolocs: My previous comment was not thought through.

You don't have to return the channel. 

```go
logCh := make(chan *loggingpb.LogEntry)

// This func call will be blocking.
if err := p.StreamPull(ctx, filter, logCh); err != nil {
  // handle err
}

// Only close the channel once this func returns,
// and no more logs will be sent to the channel.
close(logCh)
```

pkg/cloudlogging/puller_test.go
```diff
@@ -98,13 +98,119 @@ func TestPull(t *testing.T) {
...
+	}
+
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			ch := make(chan *loggingpb.LogEntry)
+			var logCh chan<- *loggingpb.LogEntry
```
yolocs: You should only need one channel for the logs

pkg/cloudlogging/puller_test.go
```diff
@@ -98,13 +98,119 @@ func TestPull(t *testing.T) {
...
+				ctx,
+				fakeClient,
+				testResource,
+				WithRetry(retry.WithMaxRetries(0, retry.NewFibonacci(500*time.Millisecond))),
+			)
+
+			var gotPullErr error
+			go func() {
+				logCh, gotPullErr = p.StreamPull(ctx, tc.filter, ch, streamClient)
+				defer close(logCh)
```
yolocs: Like what I said in the other comment, you don't need to return the channel.

```suggestion
        defer close(logCh) // If StreamPull returns, we can safely close the channel
				gotPullErr := p.StreamPull(ctx, tc.filter, logCh)
```

pkg/cloudlogging/puller.go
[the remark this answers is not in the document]
sailorlqh: Moved channel close back to StreamPull

pkg/cloudlogging/puller_test.go
```diff
@@ -115,23 +221,62 @@ func setupFakeClient(t *testing.T, ctx context.Context, s *fakeServer) *logging.
...
+
 type fakeServer struct {
 	loggingpb.UnimplementedLoggingServiceV2Server
 
-	req         *loggingpb.ListLogEntriesRequest
-	resp        *loggingpb.ListLogEntriesResponse
+	listReq     *loggingpb.ListLogEntriesRequest
+	listResp    *loggingpb.ListLogEntriesResponse
+	tailReq     *loggingpb.TailLogEntriesRequest
+	tailResp    *loggingpb.TailLogEntriesResponse
```
yolocs, line 255: It might make sense to have a counter and each tail call we just send one log back and increase the counter until a cap

yolocs (changes requested):

pkg/cloudlogging/puller.go
sailorlqh: I thought the caller should provide the client, so it can close the client by itself, therefore StreamPull don't need to handle the close of the client?

pkg/cloudlogging/puller_test.go
sailorlqh: This is removed in the latest commit.

pkg/cloudlogging/puller.go
yolocs: The stream client should be generated by calling logging client in this func. With that, it's OK to close the stream client here (you create it and close it in the same scope).

sqin2019: How about making the Puller struct to have a stream client.

pkg/cloudlogging/puller.go
sailorlqh: should we use a go routine to do the p.StreamPull?

yolocs: In test, yes. In the real command, no.

pkg/cloudlogging/puller.go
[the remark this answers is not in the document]
sailorlqh: Update.

pkg/cloudlogging/puller.go
sailorlqh: Updated to create and close stream client in streamPull

pkg/cloudlogging/puller_test.go
sailorlqh: Added

pkg/cloudlogging/puller_test.go
sailorlqh: Updated.

pkg/cloudlogging/puller_test.go
yolocs: That's not what I meant. I meant instead of providing the logs via `tailResp *loggingpb.TailLogEntriesResponse`, you can generate log on the fly. See my other comment.

pkg/cloudlogging/puller_test.go
```diff
@@ -126,12 +232,34 @@ func setupFakeClient(t *testing.T, ctx context.Context, s *fakeServer) *logging.
...
+		return fmt.Errorf("failed to receive tailLogEntry request")
+	}
+	if s.tailResp != nil {
+		// send one log entry at a time to make sure tail StreamPull can receive
+		// multiple response for server.Send()
+		for _, v := range s.tailResp.Entries {
+			s.logCounter += 1
+			if err := server.Send(&loggingpb.TailLogEntriesResponse{Entries: []*loggingpb.LogEntry{v}}); err != nil {
+				return fmt.Errorf("server failed to send: %w", err)
+			}
```
yolocs, line 261:
```suggestion
			if err := server.Send(&loggingpb.TailLogEntriesResponse{Entries: []*loggingpb.LogEntry{{LogName: fmt.Sprintf("test-%d", s.counter)}}}); err != nil {
				return fmt.Errorf("server failed to send: %w", err)
			}
			s.counter += 1
```

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,44 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+
+// SteamPull pulls live log entries, it won't stop until a cancel signal happens.
+func (p *Puller) StreamPull(ctx context.Context, filter string, logCh chan<- *loggingpb.LogEntry) error {
+	if err := retry.Do(ctx, p.retry, func(ctx context.Context) error {
+		req := &loggingpb.TailLogEntriesRequest{
+			ResourceNames: []string{p.resource},
+			Filter:        filter,
+		}
+		var tailLogEntriesClient loggingpb.LoggingServiceV2_TailLogEntriesClient
+		var createClientErr error
```
sqin2019, line 125: Why not inline?

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,44 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+		if tailLogEntriesClient, createClientErr = p.client.TailLogEntries(ctx); createClientErr != nil {
+			return fmt.Errorf("failed to create tailLogEntriesClient: %w", createClientErr)
+		}
+		defer func() {
+			_ = tailLogEntriesClient.CloseSend()
+		}()
```
sqin2019, line 131: This should be outside of retry, otherwise each retry creates a new client.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,44 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+				return retry.RetryableError(fmt.Errorf("failed to receive response: %w", err))
+			}
+			if resp.GetEntries() != nil {
+				for _, v := range resp.GetEntries() {
+					logCh <- v
+				}
+			}
+		}
+	}); err != nil {
+		fmt.Println(err)
```
sqin2019, line 152: nit: remove

sailorlqh: Done

pkg/cloudlogging/puller.go
sailorlqh: Moved to outside of retry.

pkg/cloudlogging/puller.go
```diff
@@ -33,6 +34,7 @@ import (
 // billingAccounts.
 type Puller struct {
 	client *logging.Client
+	// streamClient loggingpb.LoggingServiceV2_TailLogEntriesClient
```
yolocs, line 37: Remove?

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,44 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+				for _, v := range resp.GetEntries() {
+					logCh <- v
+				}
+			}
+		}
+	}); err != nil {
+		return fmt.Errorf("failed to pull log entries: %w", err)
+	}
+	return nil
+}
```
yolocs, line 156:
```suggestion
func (p *Puller) StreamPull(ctx context.Context, filter string, logCh chan<- *loggingpb.LogEntry) (err error) {
	tailClient, err := p.client.TailLogEntries(ctx)
	if err != nil {
		return fmt.Errorf("failed to create tailLogEntriesClient: %w", err)
	}
	defer func() {
		err = tailClient.CloseSend()
	}()

	if err := retry.Do(ctx, p.retry, func(ctx context.Context) error {
		req := &loggingpb.TailLogEntriesRequest{
			ResourceNames: []string{p.resource},
			Filter:        filter,
		}

		if err := tailLogEntriesClient.Send(req); err != nil {
			return retry.RetryableError(fmt.Errorf("failed to send request: %w", err))
		}

		for {
			resp, err := tailLogEntriesClient.Recv()
			if errors.Is(err, io.EOF) {
				continue
			}
			if err != nil {
				return retry.RetryableError(fmt.Errorf("failed to receive response: %w", err))
			}
			if resp.GetEntries() != nil {
				for _, v := range resp.GetEntries() {
					logCh <- v
				}
			}
		}
	}); err != nil {
		return fmt.Errorf("failed to pull log entries: %w", err)
	}
	return nil
}
```

yolocs: Shouldn't ignore error in the deferred func

pkg/cloudlogging/puller_test.go
```diff
@@ -98,13 +98,117 @@ func TestPull(t *testing.T) {
...
+			// create a go routine to receive logs sent from
+			// streamPull,  break after receive enough logs.
+			go func() {
+				for l := range ch {
+					gotLogs = append(gotLogs, l)
+					if len(gotLogs) == tc.wantNum {
+						done <- struct{}{}
+						break
+					}
+				}
```
yolocs, line 169: I think I found an easier way to do this. Can you try it out?

```go
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)

go func() {
  for l := range ch {
    gotLogs = append(gotLogs, l)
    if len(gotLogs) == tc.wantNum {
      cancel() // got enough logs, we can stop now
      break
    }
  }
}

// set up fake client

go func() {
  defer close(ch)
  gotPullErr = p.StreamPull(ctx, tc.filter, ch)
}

<-ctx.Done() // Either we timed out or we got enough logs and explicitly cancelled it

// Do the checks
```

pkg/cloudlogging/puller_test.go
```diff
@@ -98,13 +98,117 @@ func TestPull(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			ch := make(chan *loggingpb.LogEntry, 2)
```
yolocs, line 152:
```suggestion
			ch := make(chan *loggingpb.LogEntry)
```

pkg/cloudlogging/puller_test.go
```diff
@@ -98,13 +98,117 @@ func TestPull(t *testing.T) {
...
+			go func() {
+				defer close(ch) // If StreamPull returns, we can safely close the channel
+				gotPullErr = p.StreamPull(ctx, tc.filter, ch)
+			}()
+
+			select {
+			case <-done:
+				t.Logf("Recevied enough logEntries")
+			case <-time.After(5 * time.Second):
+				t.Logf("No enough logEntries recevied after time out")
```
yolocs, line 192:
```suggestion
				t.Logf("Not enough LogEntries received before timeout")
```

pkg/cloudlogging/puller_test.go
```diff
@@ -126,12 +230,37 @@ func setupFakeClient(t *testing.T, ctx context.Context, s *fakeServer) *logging.
 type fakeServer struct {
 	loggingpb.UnimplementedLoggingServiceV2Server
 
-	req         *loggingpb.ListLogEntriesRequest
-	resp        *loggingpb.ListLogEntriesResponse
+	listReq     *loggingpb.ListLogEntriesRequest
+	listResp    *loggingpb.ListLogEntriesResponse
+	tailReq     *loggingpb.TailLogEntriesRequest
+	logCounter  int64
```
yolocs, line 236:
```suggestion
	tailCounter  int64
```

pkg/cloudlogging/puller_test.go
```diff
@@ -126,12 +230,37 @@ func setupFakeClient(t *testing.T, ctx context.Context, s *fakeServer) *logging.
...
+	logNumCap   int64
```
yolocs, line 237: Since this is local, can we have a hardcoded cap? Like 10 should be enough

pkg/cloudlogging/puller.go
sailorlqh: Updated

pkg/cloudlogging/puller_test.go
sailorlqh: Tested and updated.

pkg/cloudlogging/puller_test.go
sailorlqh: const `fakeServerTailLogCap = 10`, and removed logNumCap from `fakeServer` field

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,49 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+		return fmt.Errorf("failed to create tailLogEntriesClient: %w", err), nil
+	}
+
+	// we don't want to ignore errors from defer but if we use
+	// err = tailClient.CloseSend()
+	// this will override errors from pulling log entries,
+	// so we use a closeTailClientErr to track errors
+	// from tailClient.CloseSend().
+	defer func() {
+		closeTailClientErr = tailClient.CloseSend()
```
yolocs, line 130: Why do we need a separate error?

pkg/cloudlogging/puller_test.go
```diff
@@ -98,9 +100,110 @@ func TestPull(t *testing.T) {
...
+			t.Parallel()
+
+			ch := make(chan *loggingpb.LogEntry)
+			var gotLogs []*loggingpb.LogEntry
+			done := make(chan struct{}, 1)
+			t.Cleanup(func() {
+				close(done)
+			})
+
+			ctx, cancel := context.WithCancel(context.Background())
```
yolocs, line 161: We need to use `context.WithTimeout`?

pkg/cloudlogging/puller_test.go
```diff
@@ -98,9 +100,110 @@ func TestPull(t *testing.T) {
...
+				testResource,
+				WithRetry(retry.WithMaxRetries(0, retry.NewFibonacci(500*time.Millisecond))),
+			)
+			var gotPullErr error
+			var gotCloseClientErr error
+
+			go func() {
+				defer close(ch)
+				gotPullErr, gotCloseClientErr = p.StreamPull(ctx, tc.filter, ch)
+				cancel()
```
yolocs, line 187: I'm not sure if this cancel is required. Can you explain?

pkg/cloudlogging/puller_test.go
```diff
@@ -33,6 +33,8 @@ import (
 
 const testResource = "test-resource"
 
+const fakeServerTailLogCap = 10
```
yolocs, line 36: nit: group consts

pkg/cloudlogging/puller.go
sailorlqh: As I explained in the comment, if there isn't separate error, pullErr in test will be nil and making the test fail.

pkg/cloudlogging/puller_test.go
sailorlqh: Sorry, removed the wrong block after testing. Updated.

pkg/cloudlogging/puller_test.go
sailorlqh: Removed.

pkg/cloudlogging/puller.go
sqin2019: We can have one retErr, and join all the errors we want to return.

sailorlqh: Used errors.Join to combine the error. Now only return one error.

yolocs: +1 the join the error

pkg/cloudlogging/puller_test.go
yolocs: I see. Can you comment?

pkg/cloudlogging/puller_test.go
```diff
@@ -98,9 +101,110 @@ func TestPull(t *testing.T) {
...
+			t.Parallel()
+
+			ch := make(chan *loggingpb.LogEntry)
+			var gotLogs []*loggingpb.LogEntry
+			done := make(chan struct{}, 1)
+			t.Cleanup(func() {
+				close(done)
+			})
+
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
yolocs, line 162: Do we need 10s? I imagine 2s would be plenty.

pkg/cloudlogging/puller_test.go
```diff
@@ -98,9 +101,110 @@ func TestPull(t *testing.T) {
...
+			// set up fake client
+			fakeClient := setupFakeClient(t, ctx, tc.server)
+			p := NewPuller(
+				ctx,
+				fakeClient,
+				testResource,
+				WithRetry(retry.WithMaxRetries(0, retry.NewFibonacci(500*time.Millisecond))),
+			)
+			var gotPullErr error
+			var gotCloseClientErr error
```
yolocs, line 183: Remove? Since we only have one error now

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,47 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+		}
+
+		if err := tailClient.Send(req); err != nil {
+			return retry.RetryableError(fmt.Errorf("failed to send request: %w", err))
+		}
+
+		for {
+			resp, err := tailClient.Recv()
+			if errors.Is(err, io.EOF) {
+				continue
```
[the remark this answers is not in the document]
sailorlqh: Haven't search the doc yet, but base on experiments, when there are no new logs to be return, and tailClient.Recv() is called, it will be an EOF

pkg/cloudlogging/puller_test.go
sailorlqh: Changed to 2s

pkg/cloudlogging/puller_test.go
sailorlqh: Removed.

pkg/cloudlogging/puller_test.go
sailorlqh: Done.

pkg/cloudlogging/puller.go
sailorlqh: Joined.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,47 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+func (p *Puller) StreamPull(ctx context.Context, filter string, logCh chan<- *loggingpb.LogEntry) error {
+	var rErr error
+	tailClient, err := p.client.TailLogEntries(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create tailLogEntriesClient: %w", err)
+	}
+
+	defer func() {
+		if closeClientErr := tailClient.CloseSend(); closeClientErr != nil {
+			rErr = errors.Join(rErr, fmt.Errorf("failed to close tailClient %w", closeClientErr))
```
sqin2019, line 127: I don't think the error is ever returned.

sqin2019: I tried in here: https://go.dev/play/p/x0pcIa73OVw

sailorlqh: Updated.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,47 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+				for _, v := range resp.GetEntries() {
+					logCh <- v
+				}
+			}
+		}
+	}); err != nil {
+		rErr = errors.Join(rErr, fmt.Errorf("failed to pull log entries: %w", err))
+		return rErr
+	}
+	return nil
```
sqin2019, line 158: Nit: an easier way is 
```
}); err != nil {
		rErr = errors.Join(rErr, fmt.Errorf("failed to pull log entries: %w", err))
	}
	return
```

pkg/cloudlogging/puller_test.go
```diff
@@ -98,9 +101,108 @@ func TestPull(t *testing.T) {
...
+
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			ch := make(chan *loggingpb.LogEntry)
+			var gotLogs []*loggingpb.LogEntry
+			done := make(chan struct{}, 1)
```
sqin2019, line 157: nit: `done := make(chan struct{})`

pkg/cloudlogging/puller.go
sailorlqh: Done

pkg/cloudlogging/puller_test.go
sailorlqh: Done.

sqin2019: test failed because of this change? On a second look, do we still need this done channel, I don't see it is used anywhere.

sailorlqh: Done is removed now, I checked, it's not used anymore, `ctx.Done()` is used instead.

sqin2019 (approved):

pkg/cloudlogging/puller.go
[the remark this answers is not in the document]
yolocs: Got it. And we rely on retry to backoff.

pkg/cloudlogging/puller.go
```diff
@@ -111,3 +113,46 @@ func (p *Puller) Pull(ctx context.Context, filter string, maxCount int) ([]*logg
...
+// SteamPull pulls live log entries, it won't stop until a cancel signal happens.
+func (p *Puller) StreamPull(ctx context.Context, filter string, logCh chan<- *loggingpb.LogEntry) (rErr error) {
+	tailClient, err := p.client.TailLogEntries(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create tailLogEntriesClient: %w", err)
+	}
+
+	defer func() {
+		if closeClientErr := tailClient.CloseSend(); closeClientErr != nil {
+			rErr = errors.Join(rErr, fmt.Errorf("failed to close tailClient %w", closeClientErr))
```
yolocs, line 126:
```suggestion
		if err := tailClient.CloseSend(); err != nil {
			rErr = errors.Join(rErr, fmt.Errorf("failed to close tailClient %w", err))
```

yolocs (approved):

yolocs (approved):

sqin2019 (approved):

PR #448, opened 2023-08-25
Add steam pull to cli

sailorlqh:


pkg/cli/tail_test.go
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !race
+// +build !race
+
```
sethvargo, line 17: This is generally not a good idea, sine it means there's a data race. What is the race error you're getting?

sethvargo (changes requested):
If we're trying to mirror the linux `tail` command, then the flag would be `-f` and `-follow`.

pkg/cli/tail.go
```diff
@@ -222,16 +238,46 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+			}
+			c.Outf(stripSpaces(string(js)))
+			if c.flagValidate {
+				if err := validation.Validate(string(js), extra...); err != nil {
+					failCount++
+					c.Errf("failed to validate log (InsertId: %q): %w\n", l.InsertId, err)
+				} else {
+					c.Outf("Successfully validated log (InsertId: %q)\n", l.InsertId)
+				}
+				c.Outf("Validation failed for %d logs (out of %d)", failCount, totalCount)
```
sqin2019, line 268: This should be printed at the end of this go routine. Probably can use defer.

pkg/cli/tail.go
```diff
@@ -222,16 +238,46 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+
+	return c.StreamTail(ctx, logCh)
+	// return err
+}
+
+func (c *TailCommand) tail(ctx context.Context) ([]*loggingpb.LogEntry, error) {
+	p, err := c.createPuller(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create logging client: %w", err)
 	}
```
sqin2019, line 281: This can be shared.

pkg/cli/tail.go
```diff
@@ -242,6 +288,30 @@ func (c *TailCommand) tail(ctx context.Context) ([]*loggingpb.LogEntry, error) {
 	return ls, nil
 }
 
+func (c *TailCommand) StreamTail(ctx context.Context, logCh chan<- *loggingpb.LogEntry) error {
```
sqin2019, line 291: We probably don't need this helper func? also tail helper func, and createPuller, it is hard to read with too many helper funcs and I don't see why we need them?

pkg/cli/tail.go
```diff
@@ -222,16 +238,46 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
-		logClient, err := logging.NewClient(ctx)
-		if err != nil {
-			return nil, fmt.Errorf("failed to create logging client: %w", err)
+func (c *TailCommand) streamTail(ctx context.Context) error {
+	var extra []validation.Validator
+	if c.flagAdditionalCheck {
+		extra = append(extra, validation.ValidateLabels)
+	}
+
+	logCh := make(chan *loggingpb.LogEntry)
```
sqin2019, line 247: logCh should be closed when go routine exits.

pkg/cli/tail.go
```diff
@@ -222,16 +238,46 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+				} else {
+					c.Outf("Successfully validated log (InsertId: %q)\n", l.InsertId)
+				}
+				c.Outf("Validation failed for %d logs (out of %d)", failCount, totalCount)
+			}
 		}
-		p = cloudlogging.NewPuller(ctx, logClient, c.flagScope)
+	}()
+
+	return c.StreamTail(ctx, logCh)
```
sqin2019, line 273: I don't see where you listen to the context cancel event here, something like 
```
	select {
	case <-ctx.Done():
		return fmt.Errorf("stream tail validate cancelled: %w", ctx.Err())
	case <-pullErr:
		return <-pullErr
	}
```

pkg/cli/tail.go
```diff
@@ -242,6 +288,30 @@ func (c *TailCommand) tail(ctx context.Context) ([]*loggingpb.LogEntry, error) {
...
+		return fmt.Errorf("failed to create logging client: %w", err)
+	}
+
+	if err := p.StreamPull(ctx, c.queryFilter(), logCh); err != nil {
+		return fmt.Errorf("StreamPull failed: %w", err)
+	}
+	return nil
+}
+
+func (c *TailCommand) createPuller(ctx context.Context) (logPuller, error) {
```
sqin2019, line 303: This can be in the Run func

sqin2019:
Did you run the command to test if it works? If not, you can go install the cmd (where your main.go is), and run the command with your test project.

pkg/cli/tail.go
```diff
@@ -61,6 +62,8 @@ type TailCommand struct {
 
 	flagAdditionalCheck bool
 
+	flagIsStream bool
```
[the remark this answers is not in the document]
sailorlqh: updated to -f

pkg/cli/tail.go
```diff
@@ -222,16 +238,46 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
-		p = c.testPuller
-	} else {
-		logClient, err := logging.NewClient(ctx)
-		if err != nil {
-			return nil, fmt.Errorf("failed to create logging client: %w", err)
+func (c *TailCommand) streamTail(ctx context.Context) error {
+	var extra []validation.Validator
+	if c.flagAdditionalCheck {
+		extra = append(extra, validation.ValidateLabels)
+	}
```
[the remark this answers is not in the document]
sailorlqh: Done

pkg/cli/tail.go
sailorlqh: Done

pkg/cli/tail.go
sailorlqh: Done

pkg/cli/tail.go
sailorlqh: Updated.

pkg/cli/tail.go
sailorlqh: Done

pkg/cli/tail.go
sailorlqh: Done

pkg/cli/tail.go
sailorlqh: Moved to Run()

pkg/cli/tail.go
```diff
@@ -242,6 +288,30 @@ func (c *TailCommand) tail(ctx context.Context) ([]*loggingpb.LogEntry, error) {
 	return ls, nil
 }
 
+func (c *TailCommand) StreamTail(ctx context.Context, logCh chan<- *loggingpb.LogEntry) error {
+	p, err := c.createPuller(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create logging client: %w", err)
+	}
+
+	if err := p.StreamPull(ctx, c.queryFilter(), logCh); err != nil {
```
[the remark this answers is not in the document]
sailorlqh: Done

pkg/cli/tail.go
```diff
@@ -222,24 +254,53 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+		close(errCh)
+	}()
 
-	return ls, nil
+	select {
+	case <-ctx.Done():
+		if c.flagValidate {
+			c.Outf("Validation failed for %d logs (out of %d)", failCount, totalCount)
+		}
+		return fmt.Errorf("stream tail validate cancelled: %w", ctx.Err())
```
sqin2019, line 297: Correct myself, maybe log and return nil instead.

pkg/cli/tail.go
```diff
@@ -222,24 +267,68 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+		}
+		close(logCh)
+		close(errCh)
+	}()
+
+	select {
+	case <-ctx.Done():
+		// main channel will block here until doneCh receives
+		// this way we can prevent the possible race condition
+		// that main channel reads from totalCount and failCount
```
sqin2019, line 318: It is not just totalCount and failCount, c.Outf and stdout.String() and etc. will also race if not blocked properly. Also we should move `if c.flagValidate {
			c.Outf("Validation failed for %d logs (out of %d)", failCount, totalCount)
		}` to the end of the read logCh goroutine. Doing so to avoid dup in line 327, they are out of the data race concern as they are only referenced there.

pkg/cli/tail.go
```diff
@@ -176,8 +189,45 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+		}
+	} else {
+		if err := c.listTail(ctx, extra, puller); err != nil {
+			rErr = errors.Join(rErr, err)
+		}
+	}
+	return
+}
+
+func (c *TailCommand) listTail(ctx context.Context, extra []validation.Validator, puller logPuller) error {
```
yolocs, line 225: nit: just name the funcs as "list" and "stream"

pkg/cli/tail.go
```diff
@@ -222,24 +267,67 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+	// doneCh is used to notifiy the main goroutinue when
+	// StreamPull is finished or canceled.
+	doneCh := make(chan struct{})
 
-	ls, err := p.Pull(ctx, c.queryFilter(), c.flagMaxNum)
-	if err != nil {
-		return nil, fmt.Errorf("failed to pull logs: %w", err)
-	}
+	var failCount int
+	var totalCount int
```
yolocs, line 280:
```suggestion
	var failCount, totalCount int
```

pkg/cli/tail.go
```diff
@@ -222,24 +267,67 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
 
-	return ls, nil
+	go func() {
+		if err := puller.StreamPull(ctx, c.queryFilter(), logCh); err != nil {
+			errCh <- fmt.Errorf("StreamPull failed: %w", err)
```
yolocs, line 284: Why do we need an error channel? Can't we do:

```go
var perr error
go func() {
	if err := puller.StreamPull(ctx, query, logCh); err != nil {
		perr = err
	}
}()
```

pkg/cli/tail.go
```diff
@@ -222,24 +267,67 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+					c.Errf("failed to validate log (InsertId: %q): %w\n", l.InsertId, err)
+				} else {
+					c.Outf("Successfully validated log (InsertId: %q)\n", l.InsertId)
+				}
+			}
+		}
+	}()
+
+	select {
+	case <-ctx.Done():
```
yolocs, line 320: Why do we care about this? Context is cancelled => StreamPull will return => close the `logCh` => exit the for loop on `logCh` => close the `doneCh`. So IIUC, you can just wait for `<-doneCh`?

pkg/cli/tail_test.go
```diff
@@ -230,3 +391,19 @@ func (p *fakePuller) Pull(ctx context.Context, filter string, maxNum int) ([]*lo
...
+		return p.injectErr
+	}
+	for _, v := range p.logEntries {
+		logCh <- v
+	}
+
+	// mock real streamPull's behavior, looping until canceled
+	// instead of using a for{} that loops forever, this is
+	// more efficient in test
+	time.Sleep(3 * time.Second)
```
yolocs, line 407: The comment doesn't explain why the sleep is necessary. Can you elaborate?

yolocs (changes requested):

pkg/cli/tail.go
sailorlqh: Updated the comment.

pkg/cli/tail.go
sailorlqh: I think return an error here is OK, as when the CLI cancel, it will be printed out anyway, have a log showing other information like timestamp and stuff isn't really that useful.

pkg/cli/tail.go
sailorlqh: Updatad, removed errCh, and just wait on doneCh.

pkg/cli/tail_test.go
sailorlqh: Comment updated.

pkg/cli/tail.go
```diff
@@ -176,8 +189,45 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+		logClient, err := logging.NewClient(ctx)
+		defer func() {
+			if err := logClient.Close(); err != nil {
+				rErr = errors.Join(rErr, fmt.Errorf("failed to close log client: %w", err))
+			}
+		}()
+		if err != nil {
+			return errors.Join(rErr, fmt.Errorf("failed to create logging client: %w", err))
+		}
+		puller = cloudlogging.NewPuller(ctx, logClient, c.flagScope, cloudlogging.WithRetry(retry.WithMaxRetries(1000000, retry.NewFibonacci(500*time.Millisecond))))
```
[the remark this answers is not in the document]
sailorlqh: Updated, now default to retry with constant to 1second, and #retry is read from cmd, and default to 1 hour.

pkg/cli/tail.go
```diff
@@ -154,6 +157,14 @@ func (c *TailCommand) Flags() *cli.FlagSet {
 			`additional lumberjack specific checks on log labels.`,
 	})
 
+	f.BoolVar(&cli.BoolVar{
+		Name:    "follow",
```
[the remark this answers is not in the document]
sqin2019: We can use linux tail default which is 10 for that.

pkg/cli/tail_test.go
sailorlqh: Changed to `-f` and `-follow`. and solved the race condition problem.

pkg/cli/tail.go
sailorlqh: Made another update, now doneCh is also removed.

pkg/cli/tail.go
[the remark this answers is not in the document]
sailorlqh: Should we change the default value to 10 as a follow up?

pkg/cli/tail.go
```diff
@@ -176,21 +200,50 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+		extra = append(extra, validation.ValidateLabels)
+	}
+
+	var puller logPuller
+	if c.testPuller != nil {
+		puller = c.testPuller
+	} else {
+		logClient, err := logging.NewClient(ctx)
+		defer func() {
+			if err := logClient.Close(); err != nil {
```
sqin2019, line 214: I wonder if it would panic, if there is err when `logging.NewClient(ctx)` as it returns nil logClient.

pkg/cli/tail.go
```diff
@@ -176,21 +200,50 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+				rErr = errors.Join(rErr, fmt.Errorf("failed to close log client: %w", err))
+			}
+		}()
+		if err != nil {
+			return errors.Join(rErr, fmt.Errorf("failed to create logging client: %w", err))
+		}
+		puller = cloudlogging.NewPuller(ctx, logClient, c.flagScope, cloudlogging.WithRetry(retry.WithMaxRetries(uint64(c.flagFollowDuration), retry.NewConstant(1000*time.Millisecond))))
```
sqin2019, line 221: Any issue with the default retry? is this for using flagFollowDuration?

pkg/cli/tail.go
```diff
@@ -154,10 +161,27 @@ func (c *TailCommand) Flags() *cli.FlagSet {
...
+		Default: false,
+		Usage:   "Whether to stream the logs as they show up.",
+	})
+
+	f.IntVar(&cli.IntVar{
+		Name:    "follow-duration",
+		Aliases: []string{"fd"},
+		Default: 3600,
+		Target:  &c.flagFollowDuration,
+		Usage: `the max time in seconds for running -f if there is no new log ` +
```
sqin2019, line 177: I am not sure if it is the right usage, if it is used in retry.WithMaxRetries, it looks like the max number of retry attempts

pkg/cli/tail.go
```diff
@@ -222,24 +275,46 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+	go func() {
+		if err := puller.StreamPull(ctx, c.queryFilter(), logCh); err != nil {
+			perr = fmt.Errorf("StreamPull failed: %w", err)
+		}
+		close(logCh)
+	}()
+
+	for l := range logCh {
+		totalCount++
+		js, err := protojson.Marshal(l)
```
sqin2019, line 294: Should we wait for a second for each read?

pkg/cli/tail.go
```diff
@@ -222,24 +275,46 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
-		return nil, fmt.Errorf("failed to pull logs: %w", err)
+	if c.flagValidate {
+		c.Outf("Validation failed for %d logs (out of %d)", failCount, totalCount)
 	}
 
-	return ls, nil
+	if perr != nil {
+		return perr
+	}
+	return fmt.Errorf("stream tail validate cancelled: %w", ctx.Err())
```
sqin2019, line 317: If feels strange to return error if it is the user exit the program.

pkg/cli/tail.go
```diff
@@ -154,10 +161,27 @@ func (c *TailCommand) Flags() *cli.FlagSet {
...
+	})
+
+	f.IntVar(&cli.IntVar{
+		Name:    "follow-duration",
+		Aliases: []string{"fd"},
+		Default: 3600,
+		Target:  &c.flagFollowDuration,
+		Usage: `the max time in seconds for running -f if there is no new log ` +
+			`to validate. Only work with -f.`,
+	})
```
yolocs, line 179: It's not clear why we need this flag

pkg/cli/tail.go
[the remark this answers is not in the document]
yolocs: Why not now?

pkg/cli/tail.go
```diff
@@ -222,24 +275,46 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+	var perr error
+	go func() {
+		if err := puller.StreamPull(ctx, c.queryFilter(), logCh); err != nil {
+			perr = fmt.Errorf("StreamPull failed: %w", err)
+		}
+		close(logCh)
+	}()
+
+	for l := range logCh {
+		totalCount++
```
yolocs, line 293: @sqin2019 Is it intentional we're ignoring max-num here?

pkg/cli/tail.go
```diff
@@ -176,21 +200,50 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+				rErr = errors.Join(rErr, fmt.Errorf("failed to close log client: %w", err))
+			}
+		}()
+		if err != nil {
+			return errors.Join(rErr, fmt.Errorf("failed to create logging client: %w", err))
+		}
+		puller = cloudlogging.NewPuller(ctx, logClient, c.flagScope, cloudlogging.WithRetry(retry.WithMaxRetries(uint64(c.flagFollowDuration), retry.NewConstant(1000*time.Millisecond))))
+	}
+
+	if c.flagFollow {
```
sqin2019, line 224: Should we do some checks to prevent flag conflicts? like when both -f and -n are provided. Also duration flag is useless in stream tail, we can add usage explanation or some other way to make that clear.

pkg/cli/tail_test.go
```diff
@@ -211,6 +227,180 @@ Validation failed for 1 logs (out of 2)
...
+Validation failed for 0 logs (out of 1)
+`, validLogJSON),
+			expErrSubstr: "stream tail validate cancelled",
+		},
+		{
+			name: "stream_tail_fail",
+			args: []string{
+				"-scope", "projects/foo",
+				"-override-filter", "test-filter",
+				"-f",
```
sqin2019, line 321: with -v, the output will be different, should we test this case as well?

pkg/cli/tail.go
yolocs, line 317: We probably shouldn't return an error here. It's expected if context is cancelled (triggered by the user or timeout), we end the stream. Probably use a debug log instead.

pkg/cli/tail_test.go
```diff
@@ -201,7 +204,20 @@ Validation failed for 1 logs (out of 2)
...
+			// flakiness. If rest of the filter string is the same, it can prove
+			// the Puller got the right filter.
+			trimedExpFilter := strings.TrimSpace(tc.expFilter)
+			trimedGotFilter := strings.TrimSpace(tc.puller.gotFilter)
+			if i := strings.Index(trimedExpFilter, "timestamp >="); i != -1 {
+				trimedExpFilter = trimedExpFilter[0:i] + trimedExpFilter[i+TimestampFilterLength:]
+			}
+			if i := strings.Index(trimedGotFilter, "timestamp >="); i != -1 {
+				trimedGotFilter = trimedGotFilter[0:i] + trimedGotFilter[i+TimestampFilterLength:]
+			}
```
yolocs, line 219: Can we add a helper func?

pkg/cli/tail_test.go
```diff
@@ -230,3 +420,20 @@ func (p *fakePuller) Pull(ctx context.Context, filter string, maxNum int) ([]*lo
...
+	for _, v := range p.logEntries {
+		logCh <- v
+	}
+
+	// mock real streamPull's behavior, looping until canceled instead of using
+	// a for{} that loops forever, this is more efficient in test in
+	// TestStreamTailCommand, the contex time out is set to 2 seconds sleep 3
+	// seconds here make sure the server is still running before the time out.
+	time.Sleep(3 * time.Second)
+	return nil
```
yolocs, line 438: Using sleep is usually frowned upon. We should avoid it if we can.

Maybe try:

```suggestion
	for _, v := range p.logEntries {
		logCh <- v
	}

	<-ctx.Done()
	return nil
```

pkg/cli/tail_test.go
```diff
@@ -230,3 +420,20 @@ func (p *fakePuller) Pull(ctx context.Context, filter string, maxNum int) ([]*lo
...
+	}
+	for _, v := range p.logEntries {
+		logCh <- v
+	}
+
+	// mock real streamPull's behavior, looping until canceled instead of using
+	// a for{} that loops forever, this is more efficient in test in
+	// TestStreamTailCommand, the contex time out is set to 2 seconds sleep 3
+	// seconds here make sure the server is still running before the time out.
+	time.Sleep(3 * time.Second)
```
sqin2019, line 437: Since we cancel the context anyway, do we still need this?

sqin2019: I checked the failed tests, it seems always 3 seconds apart, wonder if this causes that.

pkg/cli/tail_test.go
```diff
@@ -27,6 +27,9 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
+// length of `timestamp >= "2023-08-29T17:32:25Z"`.
+const TimestampFilterLength = 35
```
yolocs, line 31: Do not expose

yolocs (changes requested):
The new StreamPull logic is much cleaner. A few nits and one question about the flag.

pkg/cli/tail.go
sqin2019: Should we not? I was thinking a different behavior as mentioned in [here](https://github.com/abcxyz/lumberjack/pull/448/files#r1309449205). Do you recommend to use max-num for both tail and streamtail?

sqin2019: ah... Maybe we should consider maxnum here since linux tail does.

yolocs: I think it's confusing that you can still specify max-num even with -f flag.

pkg/cli/tail.go
sqin2019: Just update the usage explanation for duration flag that it does not change the outcome if it is stream tail.

pkg/cli/tail.go
sqin2019: @yolocs what do you think?

pkg/cli/tail_test.go
sailorlqh: I think it's a `stream_tail_validate_fail` case covers this?

pkg/cli/tail_test.go
sailorlqh: It doesn't appears on my laptop, and now the list command also fails sometimes, I think trim the TimeStamp will help fix the flaky test.

pkg/cli/tail_test.go
sailorlqh: Done

pkg/cli/tail.go
sailorlqh: Now use logger.DebugContext() to log this.

pkg/cli/tail.go
sailorlqh: Removed.

pkg/cli/tail.go
sailorlqh: Updated the usage for flag `-n` and `-d`

pkg/cli/tail.go
sailorlqh: now log this instead

pkg/cli/tail.go
sailorlqh: The default retry ends ofter 5 retries, and stream will do a retry if client failed to send request, base on experiments, it's very easy to reach the limit.
I have updated the code so `list` still uses the default retry, but `stream` uses another retry.

pkg/cli/tail.go
sailorlqh: Updated the code to check if `logClient == nil` before calling `logClient.Close()`

pkg/cli/tail.go
```diff
@@ -222,24 +279,48 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+	if c.flagValidate {
+		c.Outf("Validation failed for %d logs (out of %d)", failCount, totalCount)
 	}
 
-	return ls, nil
+	if perr != nil {
+		return perr
+	}
+	log := logger.FromContext(ctx)
+	log.DebugContext(ctx, fmt.Sprintf("stream tail validate cancelled: %v", ctx.Err()))
```
sqin2019, line 322: nit: not sure, maybe `tail cancelled: %v` since we dont always do validate?

pkg/cli/tail_test.go
```diff
@@ -27,6 +27,9 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
+// length of `timestamp >= "2023-08-29T17:32:25Z"`.
+const timestampFilterLength = 35
```
sqin2019, line 31: nit: safer to use len(fmt.Sprintf("timestamp >= %q", time.Now().UTC().Format(time.RFC3339)))? or make the original cutoff global var as well.
var (
  cutoff = "timestamp >= %q"
  timestampFilterLength = len(fmt.Sprintf(cutoff, time.Now().UTC().Format(time.RFC3339)))
)

pkg/cli/tail_test.go
```diff
@@ -211,6 +223,199 @@ Validation failed for 1 logs (out of 2)
...
+
+			trimedExpFilter := strings.TrimSpace(tc.expFilter)
+			trimedGotFilter := strings.TrimSpace(tc.puller.gotFilter)
+			// Tests can be flaky since there might be a delay between
+			// calculating timestamp in test and calculating timestamp in
+			// queryFilter. So we remove the timestamp part to reduce this
+			// flakiness. If rest of the filter string is the same, it can prove
+			// the Puller got the right filter.
+			trimedExpFilter = testTrimFilterTS(t, trimedExpFilter)
+			trimedGotFilter = testTrimFilterTS(t, trimedGotFilter)
```
sqin2019, line 410: nit: can we inline like trimedExpFilter = testTrimFilterTS(t, strings.TrimSpace(tc.expFilter))

pkg/cli/tail.go
```diff
@@ -90,6 +89,14 @@ Tails the latest lumberjack log filtered by additional custom log filter:
 Tails and validates (with additional check) the latest 10 lumberjack log in the last 4 hours in the scope:
 
       {{ COMMAND }} -scope "projects/foo" -max-num 10 -duration 4h -validate -additional-check
```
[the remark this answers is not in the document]
sailorlqh: Done

pkg/cli/tail.go
```diff
@@ -315,12 +310,16 @@ func (c *TailCommand) stream(ctx context.Context, extra []validation.Validator,
 		c.Outf("Validation failed for %d logs (out of %d)", failCount, totalCount)
 	}
 
-	if perr != nil {
-		return perr
+	// perr will be context.Canceled if it's timed out or user pressed ctrl+c
+	// in this case, we don't want to return an error, so I write an debug level
+	// log instead.
```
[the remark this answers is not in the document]
sailorlqh: Done

pkg/cli/tail.go
```diff
@@ -132,8 +139,7 @@ func (c *TailCommand) Flags() *cli.FlagSet {
 		Target:  &c.flagDuration,
 		Example: "4h",
 		Default: 2 * time.Hour,
-		Usage: `Log filter that determines how far back to search for log ` +
-			`entries. This flag is ignored if -follow is set`,
```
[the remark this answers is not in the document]
sailorlqh: Moved back.

pkg/cli/tail.go
```diff
@@ -77,13 +82,21 @@ Tails and validates the latest lumberjack log in the last 2 hours in the scope:
```
sqin2019, line 82: nit, another one here. -> latest 10

sailorlqh: Done

pkg/cli/tail.go
```diff
@@ -73,17 +78,25 @@ func (c *TailCommand) Help() string {
...
-Tails the latest lumberjack log filtered by additional custom log filter:
+Tails the latest 10 lumberjack log filtered by additional custom log filter:
 
       {{ COMMAND }} -scope "projects/foo" -additional-filter "resource.type = \"foo\""
 
 Tails and validates (with additional check) the latest 10 lumberjack log in the last 4 hours in the scope:
 
       {{ COMMAND }} -scope "projects/foo" -max-num 10 -duration 4h -validate -additional-check
+
+Follow and validates the latest lumberjack log :
```
sqin2019, line 93: nit: -> Streams and validates live lumberjack log?

sqin2019 (dismissed):

pkg/cli/tail.go
yolocs: Let's skip it for now.

yolocs (approved):

yolocs (dismissed):

sqin2019 (dismissed):

pkg/cli/tail.go
```diff
@@ -222,24 +274,51 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
 
-func (c *TailCommand) tail(ctx context.Context) ([]*loggingpb.LogEntry, error) {
-	var p logPuller
-	if c.testPuller != nil {
-		p = c.testPuller
-	} else {
-		logClient, err := logging.NewClient(ctx)
+func (c *TailCommand) stream(ctx context.Context, extra []validation.Validator, puller logPuller) error {
+	// logCh is used to receive logEntriesfrom StreamPull
+	logCh := make(chan *loggingpb.LogEntry)
```
sethvargo, line 279: Should we buffer this channel?

pkg/cli/tail.go
```diff
@@ -222,24 +274,51 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
+	var failCount, totalCount int
+
+	var perr error
+
+	go func() {
+		if err := puller.StreamPull(ctx, c.queryFilter(), logCh); err != nil {
+			perr = fmt.Errorf("StreamPull failed: %w", err)
+		}
+		close(logCh)
+	}()
```
sethvargo, line 290:
```suggestion
	go func() {
		defer close(logCh)
		if err := puller.StreamPull(ctx, c.queryFilter(), logCh); err != nil {
			perr = fmt.Errorf("StreamPull failed: %w", err)
		}
	}()
```

pkg/cli/tail.go
```diff
@@ -222,24 +274,51 @@ func (c *TailCommand) Run(ctx context.Context, args []string) error {
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to pull logs: %w", err)
+	if c.flagValidate {
+		c.Outf("Validation failed for %d logs (out of %d)", failCount, totalCount)
+	}
+
+	// perr will be context.Canceled if it's timed out or user pressed ctrl+c
+	// in this case, the perr is logged instead of returned.
+	if errors.Is(perr, context.Canceled) {
+		pkglogging.FromContext(ctx).DebugContext(ctx, "stream cancelled", "error", ctx.Err())
```
sethvargo, line 317: Should this log message be outside of the error type check?

pkg/cli/tail.go
yolocs: Yes, I missed this one. Even on error, we should close the channel

pkg/cli/tail.go
sailorlqh: I think it's OK, in StreamPull (https://github.com/abcxyz/lumberjack/blob/main/pkg/cloudlogging/puller.go#L148-L153), we send the logEntries one by one, and the for loop here receives them one by one, so I don't think buffer this channel doesn't make the big difference. 
Or is there any other concerns that I didn't get?

sethvargo (approved):

PR #459, opened 2023-10-13
feat: create abc template for lumberjack

capri-xiyue:
create abc template for lumberjack


drevell (approved):
Thanks for including me on this review. If you have any feedback to share on the developer experience of using the abc CLI, I'd be very interested to hear it.

capri-xiyue:
> Thanks for including me on this review. If you have any feedback to share on the developer experience of using the abc CLI, I'd be very interested to hear it.

Hi Dave, I think it will be quite useful if abc cli supports some mechanism to enable the UT to validate that the generated file is correct. Now I need to manually run it to see whether it can generate the correct content I want. And it is a little bit inconvenient for iteration if we want to make further updates.

drevell:
> > Thanks for including me on this review. If you have any feedback to share on the developer experience of using the abc CLI, I'd be very interested to hear it.
> 
> Hi Dave, I think it will be quite useful if abc cli supports some mechanism to enable the UT to validate that the generated file is correct. Now I need to manually run it to see whether it can generate the correct content I want. And it is a little bit inconvenient for iteration if we want to make further updates.

Agreed, thanks. @chloechien is currently working on the "golden test" feature right now that lets you save the "expected" output and diff of against the actual output. It's basically just makes it easier to get the output of the template and see if it's what you expected. I expect this feature to ship in about a month or so, depending on Chloe's availability.

PR #461, opened 2023-10-23
fix: async

sqin2019:


pdewilde (dismissed):

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/modules/CloudLoggingModule.java
```diff
@@ -19,28 +19,58 @@
...
     if (ConfigUtils.shouldFailClose(configuration.getLogMode())) {
       logging.setWriteSynchronicity(Synchronicity.SYNC);
     } else {
       logging.setWriteSynchronicity(Synchronicity.ASYNC);
+      // warmup request as a temporary workaround until this fixed:
+      // https://github.com/googleapis/java-logging/issues/1469.
+      try {
+        LogEntry entry =
+            LogEntry.newBuilder(StringPayload.of("warmup cloud logging"))
+                .setSeverity(Severity.INFO)
```
yolocs, line 67: Debug level?

clients/java-logger/library/src/main/java/com/abcxyz/lumberjack/auditlogclient/modules/CloudLoggingModule.java
```diff
@@ -19,28 +19,58 @@
...
+                  .setRequestByteThreshold(2560000L)
+                  .build());
     }
     Logging logging = loggingOptionsBuilder.build().getService();
     if (ConfigUtils.shouldFailClose(configuration.getLogMode())) {
       logging.setWriteSynchronicity(Synchronicity.SYNC);
     } else {
       logging.setWriteSynchronicity(Synchronicity.ASYNC);
+      // warmup request as a temporary workaround until this fixed:
+      // https://github.com/googleapis/java-logging/issues/1469.
```
yolocs, line 63: This is super weird if true. Are you able to reproduce the problem locally?

sqin2019: Seems due to cloud run cold start, will look into a fix.

sqin2019: fixed.

terraform/modules/ci/main.tf
```diff
@@ -91,6 +91,11 @@ resource "google_cloud_run_service" "cloudlogging_backend_client_services" {
   location = var.region
 
   template {
+    metadata {
+      annotations = {
+        "run.googleapis.com/cpu-throttling" = "false"
+      }
```
yolocs, line 97: Interesting finding. Likely a story worth sharing.

yolocs (approved):

terraform/modules/ci/main.tf
```diff
@@ -91,6 +91,11 @@ resource "google_cloud_run_service" "cloudlogging_backend_client_services" {
   location = var.region
 
   template {
+    metadata {
+      annotations = {
+        "run.googleapis.com/cpu-throttling" = "false"
```
pdewilde, line 96: Another option we can look into is using something like micronaut for faster JVM startups. Really depends on what deployment patterns we expect. Spring is more popular but has a much more expensive cold start for serverless workflows.

sqin2019: Good to know, could be a topic in the future if we want to build java apps! Lumberjack creates a java client using Guice, it does not run a java app, the java app running here is a test app just for CI, should not be a concern.

PR #462, opened 2023-10-23
chore: update go client readme

sqin2019:
with `go get -u github.com/abcxyz/lumberjack/clients/go` I will get an outdated version `github.com/abcxyz/lumberjack/clients/go v0.0.0-20220923211834-05145c811069`, I guess it is because we only have one mod file and we should use `github.com/abcxyz/lumberjack` module?

### DIFF
--- a/clients/go/README.md
+++ b/clients/go/README.md
@@ -7,7 +7,7 @@ For now, to use the module:
 1.  Get the module
 
     ```
-    go get -u github.com/abcxyz/lumberjack/clients/go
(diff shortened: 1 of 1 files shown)


yolocs:
Thanks! For context, we used to have two Go modules. One at the root and one under `clients/go`. And later we "merged" them so now we only have the one under root. That's why `go get -u github.com/abcxyz/lumberjack/clients/go` was possible in the first place.

yolocs (approved):

PR #470, opened 2024-01-05
Drop bqclient

sethvargo:
bqclient is not providing significant value for this test function, and it introduces many unnecessary layers of indirection.


yolocs (approved):
I vaguely recall we did this because in pmap we needed a similar function to check things in BQ. Do we want to change that as well?

sethvargo:
@yolocs working on it 😄 - can you re-approve? I realized I missed a retry case.

yolocs (approved):

Issue #480, opened 2024-04-08
Provide a helper to tell if it's an lumberjack error


yolocs:
### TL;DR

gRPC interceptor should return a specific error type so that the caller could assert whether it's a lumberjack error

PR #481, opened 2024-04-08
Add helper func to tell if its an interceptor error

yolocs:
Fix: https://github.com/abcxyz/lumberjack/issues/480

### DIFF
--- a/clients/go/pkg/audit/interceptor.go
+++ b/clients/go/pkg/audit/interceptor.go
@@ -37,35 +37,13 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	api "github.com/abcxyz/lumberjack/clients/go/apis/v1alpha1"
+	"github.com/abcxyz/lumberjack/clients/go/pkg/auditerrors"
 	"github.com/abcxyz/lumberjack/clients/go/pkg/justification"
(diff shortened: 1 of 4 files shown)


sqin2019 (approved):

clients/go/pkg/audit/interceptor.go
```diff
@@ -44,6 +44,29 @@ import (
...
+
+// IsInterceptorErr returns true if the grpc status error is an Lumberjack
+// interceptor error.
+func IsInterceptorErr(err error) bool {
+	// Because these errors are grpc status error, we can't assert their error type.
+	// We can only check the error code and message.
+	if status.Code(err) != codes.Internal && status.Code(err) != codes.FailedPrecondition {
+		return false
+	}
+	return strings.Contains(err.Error(), interceptorErrPrefix)
```
sethvargo, line 59: I think it'd be better if we unwrapped and checked for the prefix instead of a substring?

sethvargo (commented):
Excuse my naive question, but is there no better way for custom errors in grpc?

clients/go/pkg/audit/interceptor.go
yolocs: Yes, switched to a better way.

PR #493, opened 2025-01-16
Fix Dockerfile Go version

sethvargo:


drevell (approved):

sethvargo:
I really wish there was a way to tell Docker to use the go.mod version...

PR #494, opened 2025-01-30
Updating springboot depdency to the latest version

mattrandallbecker:


pdewilde:
I think we need to increment:

  <groupId>com.abcxyz.lumberjack</groupId>
  <version>0.0.1</version>
  
  unless that is done by CI/CD stuff, I'll take a look

pdewilde:
@lock14 let me know if I'm wrong, but I think clients/java-logger/pom.xml should also have version incremented?

Its the pom that contains all the children, including shell.

pdewilde:
https://github.com/abcxyz/lumberjack/blob/main/.github/workflows/ci.yml#L62

Probably should change java lint version to be 17  as well

pdewilde:
https://github.com/abcxyz/lumberjack/blob/main/.github/workflows/ci.yml#L80

Need to update java version in setup-java for ci/cd

pdewilde:
Though if some of the project is 17 and some is 11, that may be problematic

mattrandallbecker:
> @lock14 let me know if I'm wrong, but I think clients/java-logger/pom.xml should also have version incremented?
> 
> Its the pom that contains all the children, including shell.

Got an error when I incremented the version 

```
Could not find artifact com.abcxyz.lumberjack:audit-client-library:jar:0.0.2 in central (https://repo.maven.apache.org/maven2)
```

lock14:
> @lock14 let me know if I'm wrong, but I think clients/java-logger/pom.xml should also have version incremented?
> 
> Its the pom that contains all the children, including shell.

Are we trying to do a release, or just update the dependencies?

mattrandallbecker:
Brian's competing PR fixes the issues, bless IntelliJ

PR #496, opened 2025-01-31
Removing Java integration tests

mattrandallbecker:
The Java integration tests take a dependency on springboot that has a transient dependency with an active CVE. Updating to the latest springboot requires updating the Java code from 11 to 17, which then throws internal server errors upon testing due to the dependency on the JVS client that is Java 11.

Given the difficulty of updating the entirety of Lumberjack and JVS to be Java17 compliant, it makes more sense to remove these integration tests as they are not representative of how customers use Lumberjack, no current customers use JVS.

This should be the first of future cleanup efforts to streamline the Lumberjack codebase into a version that more closely represents how it's used.


clients/java-logger/pom.xml
```diff
@@ -8,10 +8,4 @@
   <artifactId>audit-client</artifactId>
   <packaging>pom</packaging>
   <version>0.0.1</version>
-
-  <modules>
-    <module>library</module>
-    <module>shell</module>
-    <module>grpc-test-app</module>
-  </modules>
```
lock14, line 16: You still need this block. Just delete the entry for `shell`

The new modules block would be:

```
<modules>
  <module>library</module>
  <module>grpc-test-app</module>
</modules>
```

lock14 (changes requested):

clients/java-logger/pom.xml
mattrandallbecker: Done

lock14 (approved):

PR #498, opened 2025-02-14
Fix broken build, use t.Context()

sethvargo:


scripts/server.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM golang:1.23 AS builder
+FROM --platform=$BUILDPLATFORM golang:latest AS builder
```
pjh, line 1: Is it a best practice to pin the dependency version for production / release artifacts? Should we just use golang:1.24 here?

I'm fine with :latest for the test images.

pjh (approved):

scripts/server.dockerfile
sethvargo: Only now, and only for Go :). Because we declare the toolchain, it will always build with the correct Go version.

PR #502, opened 2025-03-19
cleanup: delete golang log server

capri-xiyue:
delete unused golang server


capri-xiyue:
Hi @mattrandallbecker, I tried to delete the unused golang log server but actually I found it was used in the grpc interceptor test. The client may use grpc interceptor. See https://github.com/abcxyz/lumberjack/blob/9215dcb37eddfc43c8a6a70fa81ce54f0a3b74c0/clients/go/pkg/audit/interceptor_test.go#L1104-L1115. Missing enough test converage may cause issue

PR #503, opened 2025-03-19
cleanup: removed golang server

capri-xiyue:


pjh (changes requested):
Can you please update the PR description to explain why we're deleting these files?

I don't understand why the PR title is "removed golang server", but there are Java files also being removed and not all Go files are being removed.

capri-xiyue:
Just a draft, will close it.

